PR #10, opened 2020-10-17
Use matplotlib if mayavi is not available

alexfikl:
Produces something like [this](https://i.imgur.com/1bB0maE.png).

### DIFF
--- a/modepy/tools.py
+++ b/modepy/tools.py
@@ -474,8 +474,11 @@ def estimate_lebesgue_constant(n, nodes, visualize=False):
     lebesgue_worst = np.sum(np.abs(eq_to_out), axis=1)
     lebesgue_constant = np.max(lebesgue_worst)
 
-    if visualize and dims == 2:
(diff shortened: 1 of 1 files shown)


inducer (commented):
Thanks! Could you rewrite history to get the .png out? Otherwise LGTM.

modepy/tools.py
```diff
@@ -474,23 +474,43 @@ def estimate_lebesgue_constant(n, nodes, visualize=False):
     lebesgue_worst = np.sum(np.abs(eq_to_out), axis=1)
     lebesgue_constant = np.max(lebesgue_worst)
 
-    if visualize:
+    if visualize and dims == 2:
```
alexfikl, line 477: Should this raise if visualize is `True` for other dimensions?

inducer: Yeah, probably.

inducer:
Thanks!

PR #11, opened 2020-10-21
Fix Chebyshev modes

alexfikl:


doc/conf.py
```diff
@@ -272,4 +272,4 @@
         "https://documen.tician.de/pytools": None,
         }
 
-autoclass_content = "both"
+autoclass_content = "class"
```
alexfikl, line 275: Note to self: I feel like this might have broken something, so need to check.

inducer: I agree that `class` is the desired setting, FWIW.

modepy/modes.py
```diff
@@ -126,9 +126,11 @@ def jacobi(alpha, beta, n, x):
     PL = np.zeros((Nx, n+1), np.float64)  # noqa: N806
 
     # Initial values P_0(x) and P_1(x)
-    gamma0 = (2**(alpha+beta+1)
-            / (alpha+beta+1.)
-            * gamma(alpha+1) * gamma(beta+1) / gamma(alpha+beta+1))
+    if abs(alpha + beta + 1) < 1.0e-14:
+        gamma0 = np.pi
```
inducer, line 130: Could you add a comment on what's wrong below without this conditional?

(Maybe also mention that this case covers Chebyshev of the first kind?)

modepy/quadrature/jacobi_gauss.py
```diff
@@ -126,30 +141,28 @@ def b(n):
...
 
 
 class ChebyshevGaussQuadrature(JacobiGaussQuadrature):
-    """Chebyshev-Gauss quadrature of the first/second kind are special cases
-    of Gauss–Jacobi quadrature with α = β = -0.5/0.5.
+    r"""A Gauss quadrature rule with a square root weight.
+
+    The Chebyshev-Gauss quadrature rules of the first kind and second kind
+    correspond to Gauss-Jacobi quadrature rules with
+    :math:`\alpha = \beta = 0.5` and :math:`\alpha = \beta = 0.5`,
```
inducer, line 159:
```suggestion
    :math:`\alpha = \beta = -0.5` and :math:`\alpha = \beta = 0.5`,
```

modepy/quadrature/jacobi_gauss.py
```diff
@@ -126,30 +141,28 @@ def b(n):
...
 
-    def __init__(self, N, backend=None):  # noqa
+    def __init__(self, N, backend=None):  # noqa: N803
         JacobiGaussQuadrature.__init__(self, 0, 0, N, backend)
 
 
 class ChebyshevGaussQuadrature(JacobiGaussQuadrature):
-    """Chebyshev-Gauss quadrature of the first/second kind are special cases
-    of Gauss–Jacobi quadrature with α = β = -0.5/0.5.
+    r"""A Gauss quadrature rule with a square root weight.
```
inducer, line 155:
```suggestion
    r"""A Gauss quadrature rule with weight :math:`\sqrt{1-x^2}^{\mp 1}`.
```

test/test_modes.py
```diff
@@ -37,37 +42,25 @@
...
-            if i == j:
-                true_result = 1
-            else:
-                true_result = 0
             err = abs(result-true_result)
             maxerr = max(maxerr, err)
-            if abs(result-true_result) > ebound:
-                print(("bad", alpha, beta, i, j, abs(result-true_result)))
+            if abs(result - true_result) > ebound:
+                logger.info("[FAILED] (%g, %g): (%d, %d) error %.5e",
```
inducer, line 62:
```suggestion
                logger.error("[FAILED] (%g, %g): (%d, %d) error %.5e",
```

test/test_modes.py
```diff
@@ -37,37 +42,25 @@
...
     from functools import partial
-    jac_f = [partial(jacobi, alpha, beta, n) for n in range(max_n)]
-    wf = WeightFunction(alpha, beta)
+    jac_f = [partial(m.jacobi, alpha, beta, n) for n in range(max_n)]
     maxerr = 0
 
     for i, fi in enumerate(jac_f):
         for j, fj in enumerate(jac_f):
-            result = quad(lambda x: wf(x)*fi(x)*fj(x))
+            result = quad(lambda x: fi(x)*fj(x))
```
inducer, line 56: Yikes! How did this ever work correctly in the non-Legendre case?

inducer (commented):
Thanks!

modepy/modes.py
alexfikl: Fixed in d21df36. Also extended the formula so that it actually works for all `alpha + beta == 1` cases, not just Chebyshev.

doc/conf.py
alexfikl: Looked through it and fixed some issues in cc5873b. Should be ok now.

inducer:
Thanks!

PR #13, opened 2020-10-26
Add gradient for tensor product basis

alexfikl:


modepy/tools.py
```diff
@@ -397,6 +400,42 @@ def try_add_tet(d1, d2, d3, d4):
 submesh = MovedFunctionDeprecationWrapper(simplex_submesh)
 
 
+def hypercube_submesh(node_tuples):
```
alexfikl, line 403: Shamelessly copied from [meshmode](https://github.com/inducer/meshmode/blob/master/meshmode/mesh/tools.py#L54).

inducer (approved):
Thanks! LGTM, just one question. :)

modepy/modes.py
```diff
@@ -648,6 +662,41 @@ def tensor_product_basis(dims, basis_1d):
...
+    func = (basis_1d, grad_basis_1d)
+    return tuple(
+            _TensorProductGradientBasisFunction(order, [
+                [func[i][k] for i, k in zip(iderivative, order)]
+                for iderivative in wandering_element(dims)
+                ])
+            for order in gnitb(len(basis_1d), dims))
+
+
+def legendre_tensor_product_basis(dims, order):
```
alexfikl, line 688: Should these guys be exported in `modepy.__init__`? I mostly added them for testing to match the interface of `simplex_onb`.

modepy/tools.py
alexfikl: Started bumping versions in eaa8500.

What do you think of the rename? Is it ok?

modepy/modes.py
inducer: I think they're common enough that that makes sense.

modepy/tools.py
inducer: Rename is fine.

modepy/modes.py
```diff
@@ -648,6 +663,41 @@ def tensor_product_basis(dims, basis_1d):
...
 
+
+def grad_tensor_product_basis(dims, basis_1d, grad_basis_1d):
+    """Provides derivatives for each of the basis functions generated by
+    :func:`tensor_product_basis`.
+
+    :returns: a :class:`tuple` of callables, where each one returns a
+        *dims*-dimensional :class:`tuple`, one for each derivative.
+
+    .. versionadded: 2020.2
```
inducer, line 674:
```suggestion
    .. versionadded:: 2020.2
```
Does it render properly with just one colon?

alexfikl: Nope, it was just a typo!

modepy/modes.py
alexfikl: Done in 47a416f.

inducer:
Thanks for working on this!

PR #14, opened 2020-10-29
Add quadrature on the square and cube

alexfikl:
This adds the quadrature rules from [Witherden and Vincent](https://doi.org/10.1016/j.camwa.2015.03.017).

I also added an extractor for the rules from [Festa and Sommariva](https://doi.org/10.1016/j.cam.2012.05.021) that can be found [here](https://www.math.unipd.it/~alvise/sets.html) before realizing those are GPL licensed. They're also just for the square (not the cube), but have orders up to 55 while WV only goes to 20.


modepy/tools.py
```diff
@@ -101,6 +101,13 @@ def simplex_integral(self):
                 * reduce(mul, (factorial(alpha) for alpha in self.exponents))
                 / factorial(len(self.exponents)+sum(self.exponents)))
 
+    def hypercube_integral(self):
+        """Integral over the unit hypercube."""
```
inducer, line 105:
```suggestion
        """Integral over :math:`[0,1]^n`."""
```

modepy/quadrature/witherden_vincent_quad_data.py
```diff
@@ -0,0 +1,2106 @@
+# GENERATED by modepy/contrib/weights-from-witherden-vincent.py
```
inducer, line 1: Maybe generate a copyright/license statement here?

contrib/weights-from-festa-sommariva.py
```diff
@@ -0,0 +1,116 @@
...
+    return result
+
+
+square_data = process_rules(%s)"""
+
+
+def generate_festa_sommariva_quadrature_rules(outfile, pointset="set_amr_square"):
+    filename = f"{pointset}.m"
+    download_from_web_if_not_present(
+            url=f"https://www.math.unipd.it/~alvise/POINTSETS/{filename}",
```
inducer, line 59: Ask https://archive.org to archive this and use the resulting URL.

modepy/tools.py
inducer: If you're feeling kind, maybe also define the "unit simplex" above.

inducer (commented):
Looks great. Just a couple wrinkles.

contrib/weights-from-festa-sommariva.py
alexfikl: Fixed in 1e73044 (hopefully that's what you meant)

modepy/tools.py
alexfikl: Fixed in a4363b0.

Is there a more compact expression for the simplex?

modepy/quadrature/witherden_vincent_quad_data.py
alexfikl: For the original authors? How would that look like?

modepy/tools.py
inducer: Thx! Not that I know of.

modepy/quadrature/witherden_vincent_quad_data.py
inducer:
```py
__copyright__ = "(C) Witherden and company"
__license__ = """
The glorious text of the applicable CC license
"""
```
?

alexfikl: How does 73d718f look?  There doesn't seem to be a standard header for CC licenses.

inducer:
Thanks!

PR #15, opened 2020-10-30
Add padua points and some helpers

alexfikl:
This mostly just adds the nodes for arbitrary orders (see [wiki](https://en.wikipedia.org/wiki/Padua_points)).

Some questions / remarks:
- [ ] What would be some cool things to check?
- [ ] These points are not very symmetric. In particular, they have different numbers of nodes on faces. Will the face mass matrices still work?
- [ ] For the standard case, there's also an associated Chebychev-Gauss-type quadrature. Worth adding?
- [ ] The order of the nodes is a bit haphazard at the moment. Would be useful to sort them in lexicographic order like the tensor product case.


inducer:
> These points are not very symmetric. In particular, they have different numbers of nodes on faces. Will the face mass matrices still work?

Hmm. Probably. There's a face homogeneity assumption there currently which we would need to break. Fortunately, the data structures can take this. It's just the logic that needs some help.

PR #16, opened 2020-11-04
Implement the face mass matrices for tensor products

alexfikl:
A few of things are added here:

* a new module `modepy.shapes` to define the various supported reference elements, for now simplices and hypercubes.
* use these to abstract the different behavior when constructing face matrices.
* some non-trivial moving around of the docs

The version was bumped to `2020.3` too.

- [ ] Test against unmodified meshmode: https://github.com/inducer/meshmode/pull/95
- [ ] Meshmode changes that depend on this: https://github.com/inducer/meshmode/pull/70


modepy/matrices.py
```diff
@@ -252,33 +252,68 @@ def __init__(self, face_vertices):
...
+        """
+        :arg face_vertices: an array of shape ``[dim, npts]``, where *npts*
+            should equal `2**(dim - 1)`.
+        """
+        vol_dim, npts = face_vertices.shape
+        if npts != 2**(vol_dim-1):
+            raise ValueError("face_vertices has wrong shape")
+
+        self.origin = face_vertices[:, 0].reshape(-1, 1)
+        self.span = face_vertices[:, -2:0:-1] - self.origin
```
alexfikl, line 275: Had to reverse things here too for it to work! (see inducer/meshmode#70)

modepy/matrices.py
```diff
@@ -252,33 +252,68 @@ def __init__(self, face_vertices):
...
+        return self.origin + np.einsum("ad,dn->an", self.span, points*0.5 + 0.5)
+
+
+def modal_face_mass_matrix(trial_basis, order, face_vertices,
+        test_basis=None, domain=None):
     """
     :arg face_vertices: an array of shape ``[dim, npts]``, where *npts*
         should equal *dim*.
+    :arg domain: identifier for the reference element, can be one of
+        `"simplex"` or `"hypercube"`.
```
alexfikl, line 289: I used this "domain" thing before, but I'm not exactly sure it's a good API. Any thoughts/suggestions?

modepy/matrices.py
```diff
@@ -243,42 +243,80 @@ def mass_matrix(basis, nodes):
...
 
     .. versionadded :: 2016.1
+
+    .. versionchanged:: 2020.5
```
alexfikl, line 296: Note to self: change this!

modepy/matrices.py
alexfikl: Nevermind, the tests do pass without reversing, but it does flip the vertices in the same way.

modepy/matrices.py
```diff
@@ -243,42 +243,76 @@ def mass_matrix(basis, nodes):
...
+
+    if domain == "simplex":
+        from modepy.quadrature.grundmann_moeller import \
+                GrundmannMoellerSimplexQuadrature
+        fmap = _SimplexFaceMap(face_vertices)
+        quad = GrundmannMoellerSimplexQuadrature(order, fmap.face_dim)
+    elif domain == "hypercube":
+        from modepy.quadrature import LegendreGaussTensorProductQuadrature
+        fmap = _HypercubeFaceMap(face_vertices)
+        quad = LegendreGaussTensorProductQuadrature(fmap.face_dim, order)
```
inducer, line 311: I can't help but spot that these quadrature constructors are inconsistent in arg order. Flip? (I noticed that LGTP is being added below: just in time, phew!)

modepy/matrices.py
inducer: How about this: Make `Hypercube` and `Simplex` classes somewhere that know their dimension (but nothing else) (`modepy.shapes`?), and then make `get_face_map` and `get_quadrature_with_order` and use `functools.singledispatch` to register implementations? Over time, we could use this pattern in more of modepy. Julia: the good parts. :)

modepy/modes.py
```diff
@@ -658,6 +658,10 @@ def tensor_product_basis(dims, basis_1d):
 
     .. versionadded:: 2017.1
     """
+    if dims == 0:
+        # NOTE: using to maintain consistency in the 0d case
+        return simplex_onb(dims, len(basis_1d))
```
inducer, line 663: This yuck has to go. Part of why I'm eager to have a new API (the `shapes` thing provides a convenient excuse) is to get rid of this. (Another motivation for a new API is to use symbolic basis functions.)

(Feel free to resolve this without action, but I had to say it.)

modepy/modes.py
```diff
@@ -673,6 +677,10 @@ def grad_tensor_product_basis(dims, basis_1d, grad_basis_1d):
 
     .. versionadded:: 2020.2
     """
+    if dims == 0:
+        # NOTE: using to maintain consistency in the 0d case
+        return grad_simplex_onb(dims, len(basis_1d))
```
inducer, line 682: :vomiting_face:

modepy/nodes.py
```diff
@@ -302,6 +302,10 @@ def tensor_product_nodes(dims, nodes_1d):
 
     .. versionadded:: 2017.1
     """
+    if dims == 0:
+        # NOTE: using this to maintain consistency in the 0d case
+        return warp_and_blend_nodes(dims, 1)
+
```
inducer, line 308: :hankey:

modepy/quadrature/__init__.py
```diff
@@ -82,3 +82,29 @@ def __init__(self, quad, left, right):
         Quadrature.__init__(self,
                 left + (quad.nodes+1) / 2 * length,
                 quad.weights * half_length)
+
+
+class TensorProductQuadrature(Quadrature):
+    def __init__(self, dims, quad):
```
inducer, line 88: Arg order: see above.

modepy/quadrature/__init__.py
```diff
@@ -82,3 +82,29 @@ def __init__(self, quad, left, right):
...
+                dtype=np.float,
+                count=quad.weights.size**dims)
+        assert w.size == x.shape[1]
+
+        super().__init__(x, w)
+        self.exact_to = quad.exact_to
+
+
+class LegendreGaussTensorProductQuadrature(TensorProductQuadrature):
+    def __init__(self, dims, N, backend=None):      # noqa: N803
```
inducer, line 107: Arg order: see above.

modepy/tools.py
```diff
@@ -229,6 +233,15 @@ def barycentric_to_equilateral(bary):
 # }}}
 
 
+# {{{ hypercube coordinate mapping
+
+def hypercube_unit_vertices(dims):
+    from modepy.nodes import tensor_product_nodes
+    return tensor_product_nodes(dims, np.array([-1.0, 1.0])).T
```
inducer, line 240: All this stuff would do well in `modepy.shapes`.

inducer (commented):
Thanks! Some comments/ideas below. Let me know what you think.

modepy/matrices.py
alexfikl: Stubbed something out along those lines in 704e686. What do you think?

I put a `get_basis` (and friends) in there that feel like they belong in `modepy.modes` (or `modepy.nodes`, etc) where the things are actually defined. Should I move them?

modepy/matrices.py
alexfikl: Yep! Fixed somewhere in 704e686.

modepy/quadrature/__init__.py
alexfikl: This seems like it should stay like this? All the `tensor_product_basis` and `tensor_product_nodes` functions take `(dims, thing_1d)`

modepy/matrices.py
inducer: Yeah, I'd feel better about them being moved. One advantage of the `singledispatch` solution is that these functions don't need to live in the same file as the class.

modepy/matrices.py
inducer: Hmm, upon closer inspection (and as you note below): It seems that the quadrature interfaces (Grundmann-Möller, Xiao-Gimbutas, Vioreanu-Rokhlin) are actually the inconsistent one here. Almost everybody (except them) use (dims, order). Any ideas on how we could switch them over to (dims, order) in a somewhat backward-compatible way?

modepy/matrices.py
alexfikl: Moved most things out in 7acf2b0.

modepy/matrices.py
alexfikl: Nothing comes to mind. They're both positional arguments and integers, so no particular way to distinguish between them :confounded:

modepy/shapes.py
```diff
@@ -0,0 +1,547 @@
...
+
+r"""
+:mod:`modepy.shapes` provides a generic description of the supported shapes
+(i.e. reference elements).
+
+.. currentmodule:: modepy
+
+.. autoclass:: Shape
+.. autoclass:: Face
+.. autofunction:: biunit_vertices_for_shape
```
inducer, line 11: I'm tempted to push for `biunit_vertices_for_shape`. Everywhere else calls "bi-unit" just unit. No really good reason to start being inconsistent with that now, even if it's technically more precise. What do you think?

modepy/nodes.py
```diff
@@ -291,31 +324,157 @@ def warp_and_blend_nodes(dims, n, node_tuples=None):
...
+# {{{ QN
+
+@node_tuples_for_space.register(QN)
+def _(space: QN):
+    from pytools import \
+            generate_nonnegative_integer_tuples_below as gnitb
+    return tuple([
+        tp[::-1] for tp in gnitb(space.order + 1, space.spatial_dim)
+        ])
```
alexfikl, line 447: Reopening https://github.com/alexfikl/modepy/pull/1#discussion_r533931559, since I realized it might be a bit of a nasty hack.

I basically just transposed the node tuples so that getting equidistant nodes is the same in both of those cases.

modepy/shapes.py
alexfikl: Do you mean `unit_vertices_for_shape`?

From a quick grep, only `unit_to_barycentric` seems to still be using that + some documentation that can be updated, so it doesn't seem like too big of a hassle to change everything to `biunit`.

There's a lot of `unit_nodes` in `meshmode` though, so I'm leaning more towards just using `unit_vertices_for_shape`, since the coordinates are documented.

inducer: Agreed. My thought to go back to `unit` instead of `biunit` came after some meshmode hacking. I don't/no longer think it's worth it to start being inconsistent with all that code.

alexfikl: Renamed in 338e1aa.

modepy/matrices.py
inducer: Looks like we're stuck with this inconsistency. All new code should use `(dim, order)`, but I don't think there's a good reason to dredge up the quadrature getters. I've pushed in some FIXMEs at least, so that the issue doesn't propagate. a6f8939

modepy/quadrature/__init__.py
inducer: The way I'd like to get rid of this (and `tensor_product_nodes`) is by making an interface that takes a sequence of 1D nodes/quadratures. For `tensor_product_nodes`, it'll probably have to be a new name (`tensor_product_nodes_of([...])`, `tensor_product_nodes_inhom`?), with the existing function deprecated. `TensorProductQuadrature` is new enough that I wouldn't feel bad about just changing it.

Does that sound like a reasonable plan?

modepy/nodes.py
inducer: By "transposed", you mean the order reversal of the tuple? If so, then I think that's fine. I'd like to standardize on "reversed `gnitb` order" everywhere, to minimize the potential for confusion, and avoid the flips we battled in https://github.com/inducer/meshmode/pull/70. Does that sound reasonable?

inducer: FWIW, the code here looks OK to me.

alexfikl: Yes, flipped, not transposed!

Sounds good to me! Just wanted to make sure.

modepy/quadrature/__init__.py
alexfikl: Yeah, sounds good. This PR introduces the tensor product quadrature stuff, so that's fine! 

This one can probably be hacked around to keep the same name with `tensor_product_nodes(dims_or_list_of_nodes, nodes_1d=None)`? Not sure if it's worth the trouble though.

`tensor_product_nodes_from`? `make_tensor_product_nodes`?

inducer: I think I like the same-name solution the best. `isinstance(dims, Number)` is pretty bullet-proof.

alexfikl: Updated `tensor_product_nodes` in 368bd74 and `TensorProductQuadrature` in ef01010.

Let me know if it looks OK!

modepy/quadrature/__init__.py
```diff
@@ -82,3 +102,79 @@ def __init__(self, quad, left, right):
...
+            intervals, one for each dimension of the tensor product.
+        """
+
+        from modepy.nodes import tensor_product_nodes
+        x = tensor_product_nodes([quad.nodes for quad in quads])
+        w = np.prod(tensor_product_nodes([quad.weights for quad in quads]), axis=0)
+        assert w.size == x.shape[1]
+
+        super().__init__(x, w)
+        self.exact_to = min(quad.exact_to for quad in quads)
```
alexfikl, line 124: This looks false. What's the combined order?

inducer: To my mind, the real problem is that `exact_to` is not well-defined, at least not yet. I propose we define it as "exactly integrates up to *this* (summed) degree in multiple dimensions". (and record it in the docs) Under that definition, I think `min` is correct.

inducer: How come you suspect it to be wrong?

modepy/quadrature/__init__.py
```diff
@@ -82,3 +102,79 @@ def __init__(self, quad, left, right):
         Quadrature.__init__(self,
                 left + (quad.nodes+1) / 2 * length,
                 quad.weights * half_length)
+
+
+class TensorProductQuadrature(Quadrature):
```
inducer, line 107: We've got another case of "1D is special" here, where (for example) a `TensorProductQuadrature` with one entry in `quads` works differently from the 1D quadratures it takes as input.

modepy/quadrature/__init__.py
inducer: Thanks for working on that! Left a comment.

modepy/quadrature/__init__.py
alexfikl: I just chose `min` out of pessimism and then realized that it may require some actual thought.

Added some docs in 099e587.

modepy/quadrature/__init__.py
alexfikl: The only difference seems to be that `nodes` is of shape `(1, nnodes)` instead of just `nnodes`, right?

Any preference for a fix? I can add a `squeeze` in there or implement `__new__` and just return the 1D quadrature back?

inducer: I would kind of like to push in the opposite direction, of everything being dimension-explicit, even the 1D things. Of course that'll make 1D things a tiny bit more awkward, but people can squeeze pretty easily. How about introducing a flag to the 1D quadrature getters (like `force_dim_axis=False`) and deprecating the `force_dim_axis=False` case?

modepy/quadrature/__init__.py
```diff
@@ -82,3 +102,83 @@ def __init__(self, quad, left, right):
...
+    """
+    .. attribute:: exact_to
+
+        The tensor product quadrature is exact up to this summed polynomial degree.
```
inducer, line 111: Not liking this much. `exact_to` is part of the `Quadrature` base interface, and this class shouldn't change its mind on what that means. What I was proposing is to document (in `Quadrature`) that exactness means exactness on P^N of that degree (in the multi-D) case.

modepy/quadrature/__init__.py
inducer: Thanks! Commented there. Any other reason why this wouldn't live up to the definition I proposed?

modepy/quadrature/__init__.py
alexfikl: Which getters do you mean? `JacobiGaussQuadrature` and friends?

modepy/quadrature/__init__.py
alexfikl: The docs for `Quadrature.exact_to` already mention it's about the "total degree". Isn't that sufficient to imply `P^N` even in higher dimensions?

The tensor product quadrature would integrate up to a "maximum degree", right?

modepy/quadrature/__init__.py
inducer: Yeah. Everyone that produces a 1D quadrature.

modepy/quadrature/__init__.py
inducer: > The docs for `Quadrature.exact_to` already mention it's about the "total degree"

Oof. Good point. Then most of the data returned is currently incorrect. (see [here](https://en.wikipedia.org/wiki/Degree_of_a_polynomial#Extension_to_polynomials_with_two_or_more_variables))

I propose we change that wording to P^N (giving a weaker guarantee, but one that matches the data that's actually being returned).

modepy/quadrature/__init__.py
alexfikl: 877018e

Hopefully I didn't forget anyone.

modepy/quadrature/__init__.py
alexfikl: How does 10a17d5 look?

inducer: LGTM

modepy/quadrature/__init__.py
inducer: LGTM

modepy/quadrature/__init__.py
inducer: (Superseded by discussion below.)

inducer (approved):
Looking good! Two minor things, I'll deal with them.

modepy/shapes.py
```diff
@@ -0,0 +1,547 @@
...
+
+
+def _hypercube_face_to_vol_map(face_vertices: np.ndarray, p: np.ndarray):
+    dim, npoints = face_vertices.shape
+    if npoints != 2**(dim - 1):
+        raise ValueError("'face_vertices' has wrong shape")
+
+    origin = face_vertices[:, 0].reshape(-1, 1)
+
+    # works up to (and including) 3D:
```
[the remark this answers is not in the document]
inducer: 081077b

inducer:
Thanks for seeing this through!

alexfikl:
> Thanks for seeing this through!

Yeah, this was a big one! That title is not even remotely accurate anymore!

PR #17, opened 2020-11-20
Symbolic basis

inducer:
- [x] Depends on https://github.com/inducer/pymbolic/pull/26
    - [x] Point `requirements.txt` back to master once that's in
- [x] Test with meshmode: https://github.com/inducer/meshmode/pull/88
- [x] Test with grudge: https://github.com/inducer/grudge/pull/33
- [x] Test with pytential: https://github.com/inducer/pytential/pull/40

@xywei This gives you a way to access symbolic expressions for basis functions, which might help with interpolation in volumential. (Probably best to add a bit of plumbing to meshmode, rather than access modepy directly.)

@alexfikl Turns out getting the basis functions to be symbolic does not require a whole new interface after all.

### DIFF
--- a/modepy/modes.py
+++ b/modepy/modes.py
@@ -83,6 +83,10 @@ Monomials
 
 .. autofunction:: monomial(order, rst)
 .. autofunction:: grad_monomial(order, rst)
+
+Symbolic Basis Functions
+------------------------
+.. autofunction:: symbolicize_basis(order, rst)
 """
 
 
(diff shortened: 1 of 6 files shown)


modepy/modes.py
```diff
@@ -79,22 +74,52 @@
...
+Symbolic Basis Functions
+------------------------
+.. autofunction:: symbolicize_basis(order, rst)
 """
 
 
+# {{{ helpers for symbolic evaluation
+
+def _cse(expr, prefix):
+    if "pymbolic" in sys.modules:
```
alexfikl, line 96: Why not a `try.. except ImportError`?

inducer: Because that'll have an import as a side effect.

modepy/modes.py
```diff
@@ -79,22 +74,52 @@
...
 Monomials
 ---------
 
 .. autofunction:: monomial(order, rst)
-
 .. autofunction:: grad_monomial(order, rst)
+
+Symbolic Basis Functions
+------------------------
+.. autofunction:: symbolicize_basis(order, rst)
```
alexfikl, line 89:
```suggestion
.. autofunction:: symbolicize_basis
```
?

inducer: 8131827

PR #26, opened 2021-03-29
Remove unnecessary argument from "orthonormality_weight" methods

thomasgibson:
Removes an unused argument and makes the method consistent across all supported bases.


inducer:
Thanks!

PR #28, opened 2021-04-12
Document face mass matrix functions, add nodal_quad_mass_matrix_for_face, deprecate broken use of nodal_mass_matrix_for_face, bump version

inducer:
`nodal_mass_matrix_for_face` permitted a "pseudo-overintegrated" usage, where it would use pseudoinverses to find coefficients and then try to piggyback off of `modal_face_mass_matrix`. For computing face mass matrices in vanilla nodal DG (where the pseudoinverse is the actual inverse), that is OK. But for overintegrated use, this is complete nonsense (and deprecated by this PR). This introduces a new `nodal_quad_mass_matrix_for_face` that does less, but correctly.

Prompted by @thomasgibson's prodding in https://github.com/inducer/grudge/pull/62#discussion_r611879139.

- [x] :warning: Wait for #29.

cc @lukeolson

### DIFF
--- a/modepy/matrices.py
+++ b/modepy/matrices.py
@@ -56,6 +56,7 @@ where :math:`(\phi_i)_i` is the basis of functions underlying :math:`V`.
 .. autofunction:: mass_matrix
 .. autofunction:: modal_mass_matrix_for_face
 .. autofunction:: nodal_mass_matrix_for_face
+.. autofunction:: nodal_quad_mass_matrix_for_face
 
 Differentiation is also convenient to express by using :math:`V^{-1}` to
 obtain modal values and then using a Vandermonde matrix for the derivatives
@@ -247,7 +248,17 @@ def mass_matrix(basis, nodes):
 
 def modal_mass_matrix_for_face(face: Face, face_quad: Quadrature,
         trial_functions, test_functions):
-    """
+    r"""Using the quadrature *face_quad*, provide a matrix :math:`M_f` that
+    satisfies:
+
+    .. math::
+
+        \displaystyle {(M_f)}_{ij} \approx \int_F \phi_i(r) \psi_j(r) dr,
+
+    where :math:`\phi_i` are the (volume) Lagrange basis functions obtained from
+    *test_functions* at *volume_nodes*, and :math:`\psi_j` are the (surface)
+    *trial_functions*.
+
     .. versionadded:: 2020.3
     """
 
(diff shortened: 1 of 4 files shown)


thomasgibson (approved):
Thanks for updating the docstrings and adding the new face mass routine. This makes sense and I've just verified that the grudge unit tests in https://github.com/inducer/grudge/pull/62 seem happy.

There are other problems I'm trying to deal with, but I believe they are not related to the functions here. Will update in https://github.com/inducer/grudge/pull/62.

modepy/matrices.py
```diff
@@ -265,15 +276,64 @@ def modal_mass_matrix_for_face(face: Face, face_quad: Quadrature,
...
+    *test_functions* at *volume_nodes*, and :math:`w_i` and :math:`r_i` are the
+    weights and nodes from *face_quad*.
+
+    .. versionadded :: 2021.2
+    """
+    vol_vdm = vandermonde(test_functions, volume_nodes)
+
+    mapped_nodes = face.map_to_volume(face_quad.nodes)
+
+    vol_modal_mass_matrix = np.empty((len(test_functions), len(face_quad.weights)))
```
thomasgibson, line 331: Will this work if `face_quad = mp.Quadrature(np.empty((0, 1)), np.empty((0, 1)))`? I don't think this does. Thinking of the case where the face/face_quadrature is coming from a face group on a point set (dim = 0).

thomasgibson (commented):
Just a comment concerning a corner-case I'm running into with mirgecom.

modepy/matrices.py
inducer: > `mp.Quadrature(np.empty((0, 1)), np.empty((0, 1)))`

This is strange to me. It looks like you're passing a 2D array (of shape `(0, 1)` for `weights`. Weights to me are always 1D. Could you explain?

thomasgibson: Sure. So in one of the diffusion tests in mirgecom, (1D), a `QuadratureSimplexElementGroup` is passed in as a face group (`afgrp`), which has spatial dimension zero, and a quadrature rule coming from here: https://github.com/inducer/meshmode/blob/main/meshmode/discretization/poly_element.py#L275

This is the latest commit I used in the grudge branch: https://github.com/inducer/grudge/pull/62/commits/de29e293e66b920a990c80de7f25165ca396c0ae

I managed to reproduce the error by running this in the mirgecom tests: `py.test test_diffusion.py -k 'test_diffusion_accuracy' -x --pdb`. Seems to be failing only for 1D, so you can reproduce this error using the 1D test parameter: `problem = get_static_trig_var_diff(1)`

thomasgibson: I have not ruled out my own errors, but I am wondering if we're handling 0-dim facets correctly!

inducer: (Btw, when discussing errors, please include a traceback.) (You can get a usable one with `--tb=native` in pytest. Pytest's default ones are not usable.) 

I get [this](https://gist.github.com/inducer/2bf979884c444674b4bdd87afe65d2df). Is that what you're seeing?

inducer: > `mp.Quadrature(np.empty((0, 1)), np.empty((0, 1)))`

I just realized that this is in fact a [code snippet from meshmode](https://github.com/inducer/meshmode/blob/511c8af87b09d4f5a1b39b22498664807b161c49/meshmode/discretization/poly_element.py#L275). I did not know that, and it took me a long time to figure it out.

inducer: It still doesn't make any sense to me though.

thomasgibson: > > `mp.Quadrature(np.empty((0, 1)), np.empty((0, 1)))`
> 
> I just realized that this is in fact a [code snippet from meshmode](https://github.com/inducer/meshmode/blob/511c8af87b09d4f5a1b39b22498664807b161c49/meshmode/discretization/poly_element.py#L275). I did not know that, and it took me a long time to figure it out.

Yep, I linked that in my reply above. I managed to track it down to that line, as that was causing the traceback I was seeing.



> (Btw, when discussing errors, please include a traceback.) (You can get a usable one with `--tb=native` in pytest. Pytest's default ones are not usable.)
> 
> ~I get [this](https://gist.github.com/inducer/2bf979884c444674b4bdd87afe65d2df).~ (Nvm. I was on a funny loopy branch, turns out.) I get [this](https://gist.github.com/inducer/a9624228ea8de4582bc81049ae72d2be). Is that what you're seeing?

yes.

inducer: https://github.com/inducer/meshmode/pull/158

inducer: > Yep, I linked that in my reply above. I managed to track it down to that line, as that was causing the traceback I was seeing.

OIC. It would have helped if I had seen it. Totally on me. Sorry!

Unrelated, unsolicited Github advice: Before copy-pasting a Github code URL, press `y` (the `y` key). Github's Javascript will swap out the URL for one that contains the actual Github revision. That way links remain stable even when code changes.

thomasgibson: Can confirm now that all is peaceful and happy with this PR + https://github.com/inducer/meshmode/pull/158

thomasgibson (approved):

inducer:
Thanks for taking a look!

PR #32, opened 2021-04-14
Add quadrature for 0d domains

alexfikl:
Felt bad about breaking the 0d stuff before, so followed up on inducer/grudge#62 and centralized it here.

@thomasgibson

### DIFF
--- a/modepy/quadrature/__init__.py
+++ b/modepy/quadrature/__init__.py
@@ -2,6 +2,7 @@
 .. currentmodule:: modepy
 
 .. autoclass:: Quadrature
+.. autoclass:: PointQuadrature
 
 .. autofunction:: quadrature_for_space
 
@@ -85,6 +86,18 @@ class Quadrature:
         or of shape *(npts,)* for 1D quadrature.
         """
         return np.dot(self.weights, f(self.nodes))
+
+
+class PointQuadrature(Quadrature):
+    """A quadrature rule that should be used for 0d domains (i.e. points).
+
+    Inherits from :class:`Quadrature`.
+    """
+
+    def __init__(self):
+        self.nodes = np.empty((0, 1), dtype=np.float64)
+        self.weights = np.ones((1,), dtype=np.float64)
+        self.exact_to = np.inf
 
 
 class Transformed1DQuadrature(Quadrature):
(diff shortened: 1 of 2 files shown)


modepy/quadrature/__init__.py
```diff
@@ -87,6 +88,18 @@ def __call__(self, f):
...
+class PointQuadrature(Quadrature):
+    """A quadrature rule that should be used for 0d domains (i.e. points).
+
+    Inherits from :class:`Quadrature`.
+    """
+
+    def __init__(self):
+        self.nodes = np.empty((0, 1), dtype=np.float64)
+        self.weights = np.ones((1,), dtype=np.float64)
+        self.exact_to = np.inf
```
alexfikl, line 100: Preferences on `exact_to`?

inducer: I'm OK with `np.inf`.

modepy/quadrature/__init__.py
```diff
@@ -87,6 +88,18 @@ def __call__(self, f):
         return np.dot(self.weights, f(self.nodes))
 
 
+class PointQuadrature(Quadrature):
```
inducer, line 91: I'm not sure I'm loving `PointQuadrature` as a name. (Because all quadratures have to do with points somehow...) `ZeroDQuadrature`? Also awkward, I know.

inducer (commented):
Thanks for working on this!

modepy/quadrature/__init__.py
alexfikl: Renamed in b0563c1.

Yeah, not sure I have anything better. `Quadrature0D`? That looks weird too..

thomasgibson (approved):
Ah you beat me to it (already started a branch on this too)!

This looks good and I'm fine with `np.inf` as well.

modepy/quadrature/__init__.py
thomasgibson: `ZeroDimensionalQuadrature`?

inducer: > `ZeroDimensionalQuadrature`?

That works for me. (And I like it better than the prior names.)

alexfikl: Done! 354c190

inducer:
Thanks!

Issue #33, opened 2021-04-14
Should Quadrature take an `exact_to` optional argument?


thomasgibson:
Let's say we want to create a quadrature rule outside the existing classes. For example, quadrature rules corresponding to "mass-matrix quadrature" used for subclasses of `_MassMatrixQuadratureElementGroup` in `meshmode`. It's simple enough to generate the weights and corresponding nodes. However, if we want to reason about the accuracy of the resulting quadrature rule, we're hosed (bam --- `quad.exact_to` raises an `AttributeError`).

I have already alluded to this problem [here](https://github.com/inducer/grudge/pull/62#discussion_r613554929). I proposed that `mp.Quadrature` be made an abstract class to _force a contract with the code_ --- quadrature rules should have specified accuracy. However, there are exceptions to this (notably `FejerQuadrature`)

How about we allow something like: `mp.Quadrature(nodes, weights, exact_to=order)`? Is this reasonable? If so, what should the default be for cases like `FejerQuadrature`? `None`?

Or should this be absorbed into https://github.com/inducer/modepy/issues/31 somehow?

inducer:
I think the optional argument to `mp.Quadrature` is fine. Since the agreed-upon protocol for "I don't know" seems to be "raise `AttributeError`" (maybe document this while we're at it? We can do something that makes more sense with #31.), we can simply not set the attribute if it's not given. That way, we avoid imcompatibility with subclasses which have `exact_to` as a computed `@property`.

inducer:
(cc @alexfikl, since you were part of the original discussion in https://github.com/inducer/grudge/pull/62)

inducer:
(Opinions?)

thomasgibson:
I just pushed a branch implementing the quick "band-aid" solution: https://github.com/inducer/modepy/commit/cbb4ccb8956cdd0c41c69770238e4bb752ba38ae

While I think this works, do you not think it's worth patching this up a bit more nicely (since we want to revise this as a whole anyways)?

A potentially nicer way may be to have a method for `Quadrature` and do something like:
```python
@property
def exact_to(self):
    try:
        return self._exact_to    # have to rename to something different
    except AttributeError:
        raise NotImplementedError(...)
```
This is certainly not the only option. A cleaner solution would be to go through the `@abstractproperty` route and implement methods that do the right thing (return `exact_to` for known cases, and return, say, an `Exception` for others).

inducer:
Doing that breaks a healthy bunch of existing code:
```
[...]

PR #34, opened 2021-04-14
Add an optional `exact_to` argument to the Quadrature class

thomasgibson:
Does exactly what it implies! Related to #33 and https://github.com/inducer/meshmode/issues/160

### DIFF
--- a/modepy/quadrature/__init__.py
+++ b/modepy/quadrature/__init__.py
@@ -92,7 +92,7 @@ class Quadrature:
         self.weights = weights
         # TODO: May be revamped/addressed later;
         # see https://github.com/inducer/modepy/issues/31
-        if exact_to:
+        if exact_to is not None:
             self.exact_to = exact_to
 
     def __call__(self, f):
--- a/modepy/quadrature/__init__.py
+++ b/modepy/quadrature/__init__.py
@@ -112,9 +112,9 @@ class ZeroDimensionalQuadrature(Quadrature):
     """
 
     def __init__(self):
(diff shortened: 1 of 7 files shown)


alexfikl:
Depending on how #33 turns out, maybe it's worth going through the various quadrature rules and passing `exact_to` in calls to `super().__init__()`? Like [here](https://github.com/inducer/modepy/blob/8a532888b70ed738c32cf633e87cb8e5b83c9c3e/modepy/quadrature/jacobi_gauss.py#L77-L78).

thomasgibson:
Thanks @alexfikl! I am now much happier with how this is turning out! @inducer what do you think?

modepy/quadrature/__init__.py
```diff
@@ -69,14 +69,31 @@ class Quadrature:
...
+        :arg weights: an array of length *nnodes*.
+        :arg exact_to: an optional argument denoting the symmed polynomial
+            degree to which the quadrature is exact. By default, `exact_to`
+            is `None` and will *not* be set as an attribute.
+        """
         self.nodes = nodes
         self.weights = weights
+        # TODO: May be revamped/addressed later;
+        # see https://github.com/inducer/modepy/issues/31
+        if exact_to:
```
inducer, line 95:
```suggestion
        if exact_to is not None:
```
(Otherwise you'd have trouble describing the accuracy of a brilliant new quadrature rule that's exact on constants.)

inducer (commented):
Thanks! One minor wrinkle, otherwise this looks good to me.

inducer:
Thanks!

PR #36, opened 2021-04-15
Implement, test face_normal

### LINKED ISSUE #35: Reference surface normals for `modepy.Face`s
In preparation for doing entropy-stable DG, we will need to access the surface normals defined on the reference element.

My idea is to augment the [Face class](https://github.com/inducer/modepy/blob/8a532888b70ed738c32cf633e87cb8e5b83c9c3e/modepy/shapes.py#L250) with an additional attribute: `surface_normal`.

Using this issue to track discussion/question relating to this. Will likely involve https://github.com/inducer/pymbolic to generate functions which can act on arbitrary faces, defined uniquely by their vertices, to generate the appropriate normal for each face.

inducer:
Closes #35.

### DIFF
--- a/modepy/shapes.py
+++ b/modepy/shapes.py
@@ -301,8 +301,7 @@ def face_normal(face: Face, normalize=True) -> np.ndarray:
             surface_ps = span_vec
 
     if normalize:
-        area_element = np.sqrt(surface_ps.norm_squared())
-        surface_ps = surface_ps / area_element
+        surface_ps = surface_ps / np.sqrt(surface_ps.norm_squared())
 
     # Compute the normal as the dual of the surface pseudoscalar.
     return surface_ps.dual().as_vector()
--- a/modepy/shapes.py
+++ b/modepy/shapes.py
@@ -289,22 +289,23 @@ def face_normal(face: Face, normalize=True) -> np.ndarray:
         (fv,), = face_vertices
         return np.array([np.sign(fv)])
 
+    # Compute the outer product of the vectors spanning the surface, obtaining
+    # the surface pseudoscalar.
     from pymbolic.geometric_algebra import MultiVector
-    span_multivec = None
+    surface_ps = None
     for i in range(face.dim):
         span_vec = MultiVector(face_vertices[:, i+1] - face_vertices[:, 0])
-        if span_multivec is not None:
-            span_multivec = span_multivec * span_vec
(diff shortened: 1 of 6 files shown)


modepy/shapes.py
```diff
@@ -276,6 +277,31 @@ class Face:
     map_to_volume: Callable[[np.ndarray], np.ndarray]
 
 
+def face_normal(face: Face, normalize=True) -> np.ndarray:
+    """
+    .. versionadded :: 2021.2.1
+    """
```
thomasgibson, line 283: Maybe a short description of what's going on for those that might be a bit rusty with geometric algebra :sweat_smile: (me)

inducer: 2f519b9

modepy/shapes.py
```diff
@@ -276,6 +277,37 @@ class Face:
...
+    for i in range(face.dim):
+        span_vec = MultiVector(face_vertices[:, i+1] - face_vertices[:, 0])
+        if surface_ps is not None:
+            surface_ps = surface_ps ^ span_vec
+        else:
+            surface_ps = span_vec
+
+    if normalize:
+        area_element = np.sqrt(surface_ps.norm_squared())
+        surface_ps = surface_ps / area_element
```
inducer, line 305:
```suggestion
        surface_ps = surface_ps / np.sqrt(surface_ps.norm_squared())
```

modepy/shapes.py
inducer: 84190bb is prettier.

thomasgibson: Thanks!

modepy/shapes.py
```diff
@@ -276,6 +277,34 @@ class Face:
...
+    """
+    .. versionadded :: 2021.2.1
+    """
+    volume_vertices = unit_vertices_for_shape(face.volume_shape)
+    face_vertices = volume_vertices[:, face.volume_vertex_indices]
+
+    if face.dim == 0:
+        # FIXME Grrrr. Hardcoded special case. Got a better idea?
+        (fv,), = face_vertices
+        return np.array([np.sign(fv)])
```
thomasgibson, line 290: It seems `dim == 0` likes to throw a wrench in quite a few places. However, I am fine with this personally!

thomasgibson (approved):
Thanks! This looks great

PR #37, opened 2021-04-18
Remove UNIT_VERTICES

thomasgibson:
With https://github.com/inducer/grudge/pull/62 merged, I believe it is now safe to remove `UNIT_VERTICES`. Also makes a minor change to an unnecessary alias :)

### DIFF
--- a/modepy/modal_decay.py
+++ b/modepy/modal_decay.py
@@ -78,7 +78,7 @@ def simplex_interp_error_coefficient_estimator_matrix(
 def make_mode_number_vector(mode_order_tuples, ignored_modes):
     node_cnt = len(mode_order_tuples)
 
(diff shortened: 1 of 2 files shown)


modepy/modal_decay.py
```diff
@@ -78,7 +78,7 @@ def simplex_interp_error_coefficient_estimator_matrix(
 def make_mode_number_vector(mode_order_tuples, ignored_modes):
     node_cnt = len(mode_order_tuples)
 
-    mode_number_vector = np.zeros(node_cnt-ignored_modes, dtype=np.int)
+    mode_number_vector = np.zeros(node_cnt-ignored_modes, dtype=int)
```
inducer, line 81:
```suggestion
    mode_number_vector = np.zeros(node_cnt-ignored_modes, dtype=np.int32)
```

inducer:
Thanks!

PR #41, opened 2021-08-25
Generalize tensor product spaces

alexfikl:
Introduces a `TensorProductSpace` that just takes a bunch of other `Spaces` that it then uses to construct the higher dimensional stuff in `othonormal_basis_for_space` and friends. There's also a `TensorProductShape`, but that doesn't do anything at the moment.

TODO
- [x] Add some tests
- [x] Extend `tensor_product_nodes` to take nodes with different dimensions
- [x] Extend `_TensorProductBasis` to work with nodes of different dimensions

### DIFF
--- a/modepy/__init__.py
+++ b/modepy/__init__.py
@@ -29,7 +29,7 @@ from modepy.shapes import (
         submesh_for_shape,
         )
 from modepy.spaces import (
-        FunctionSpace, PN, QN, space_for_shape)
+        FunctionSpace, TensorProductSpace, PN, QN, space_for_shape)
 from modepy.modes import (
         jacobi, grad_jacobi,
         simplex_onb, grad_simplex_onb, simplex_onb_with_mode_ids,
@@ -83,7 +83,7 @@ __all__ = [
         "face_normal", "unit_vertices_for_shape", "faces_for_shape",
         "submesh_for_shape",
 
-        "FunctionSpace", "PN", "QN", "space_for_shape",
+        "FunctionSpace", "TensorProductSpace", "PN", "QN", "space_for_shape",
 
         "jacobi", "grad_jacobi",
         "simplex_onb", "grad_simplex_onb", "simplex_onb_with_mode_ids",
--- a/modepy/__init__.py
+++ b/modepy/__init__.py
@@ -23,7 +23,7 @@ THE SOFTWARE.
 
 
 from modepy.shapes import (
-        Shape, Face, Simplex, Hypercube,
+        Shape, Face, Simplex, Hypercube, TensorProductShape,
 
         face_normal, unit_vertices_for_shape, faces_for_shape,
         submesh_for_shape,
@@ -79,7 +79,7 @@ GaussLegendreQuadrature = LegendreGaussQuadrature
 __all__ = [
         "__version__",
 
-        "Shape", "Face", "Simplex", "Hypercube",
+        "Shape", "Face", "Simplex", "Hypercube", "TensorProductShape",
         "face_normal", "unit_vertices_for_shape", "faces_for_shape",
         "submesh_for_shape",
 
--- a/modepy/modes.py
+++ b/modepy/modes.py
@@ -847,8 +847,8 @@ class Basis:
 
     def orthonormality_weight(self) -> float:
         """
-        Can raise :exc:`BasisNotOrthonormal` if the basis does not have
-        a weight.
+        :raises: :exc:`BasisNotOrthonormal` if the basis does not have
+            a weight.
         """
         raise NotImplementedError
 
--- a/modepy/modes.py
+++ b/modepy/modes.py
@@ -21,7 +21,6 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
 """
 
-
 from warnings import warn
 from math import sqrt
 from functools import singledispatch, partial
@@ -654,35 +653,68 @@ def grad_simplex_best_available_basis(dims, n):
 
 # {{{ tensor product basis helpers
 
+def _get_dim_slices(dims_per_function):
+    s = 0
+    slices = []
+    for dim in dims_per_function:
+        slices.append(np.s_[s:s + dim])
+        s += dim
+
+    return tuple(slices)
+
+
+def _handle_scalar(ary):
+    if isinstance(ary, np.ndarray):
+        ary = np.squeeze(ary)
+        if ary.shape == ():
+            ary = ary[()]
+
+    return ary
+
+
 class _TensorProductBasisFunction:
     # multi_index is here just for debugging.
 
-    def __init__(self, multi_index, per_dim_functions):
+    def __init__(self, multi_index, functions, dims_per_function=None):
+        if dims_per_function is None:
+            dims_per_function = (1,) * len(functions)
+        else:
+            assert len(dims_per_function) == len(functions)
+
         self.multi_index = multi_index
-        self.per_dim_functions = per_dim_functions
+        self.functions = functions
+        self.dims = _get_dim_slices(dims_per_function)
 
     def __call__(self, x):
         result = 1
-        for iaxis, per_dim_function in enumerate(self.per_dim_functions):
-            result = result * per_dim_function(x[iaxis])
+        for i, function in enumerate(self.functions):
+            result *= function(x[self.dims[i]])
 
-        return result
+        return _handle_scalar(result)
+
+    def __repr__(self):
+        return (f"{type(self).__name__}(mi={self.multi_index}, "
+                f"dims={self.dims}, functions={self.functions})")
 
 
 class _TensorProductGradientBasisFunction:
     # multi_index is here just for debugging.
 
-    def __init__(self, multi_index, per_dim_derivatives):
+    def __init__(self, multi_index, derivatives, dims_per_function=None):
+        if dims_per_function is None:
+            dims_per_function = (1,) * len(derivatives[0])
+
         self.multi_index = multi_index
-        self.per_dim_derivatives = tuple(per_dim_derivatives)
+        self.derivatives = tuple(derivatives)
+        self.dims = _get_dim_slices(dims_per_function)
 
     def __call__(self, x):
-        result = [1] * len(self.per_dim_derivatives)
-        for ider, per_dim_functions in enumerate(self.per_dim_derivatives):
-            for iaxis, per_dim_function in enumerate(per_dim_functions):
-                result[ider] *= per_dim_function(x[iaxis])
+        result = [1] * len(self.derivatives)
+        for ider, functions in enumerate(self.derivatives):
+            for i, function in enumerate(functions):
+                result[ider] *= function(x[self.dims[i]])
 
-        return tuple(result)
+        return tuple([_handle_scalar(r) for r in result])
 
 # }}}
 
@@ -778,7 +810,6 @@ def symbolicize_function(
     """
     import pymbolic.primitives as p
     r_sym = p.make_sym_vector(ref_coord_var_name, dim)
-
     result = f(r_sym)
 
     if dim == 1:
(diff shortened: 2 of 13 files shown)


alexfikl:
@inducer This should be ready for a first look. It doesn't seem to break the world, so hopefully there isn't anything too fishy in here.

modepy/__init__.py
```diff
@@ -79,11 +79,11 @@
...
-        "Shape", "Face", "Simplex", "Hypercube",
+        "Shape", "Face", "Simplex", "Hypercube", "TensorProductShape",
         "face_normal", "unit_vertices_for_shape", "faces_for_shape",
         "submesh_for_shape",
 
-        "FunctionSpace", "PN", "QN", "space_for_shape",
+        "FunctionSpace", "TensorProductSpace", "PN", "QN", "space_for_shape",
 
         "jacobi", "grad_jacobi",
         "simplex_onb", "grad_simplex_onb", "simplex_onb_with_mode_ids",
```
inducer, line 89: This should also fix the README and the doc index. Both only mention simplices but not tensor products.

modepy/shapes.py
```diff
@@ -559,4 +566,40 @@ def _submes_for_hypercube(shape: Hypercube, node_tuples):
 # }}}
 
 
+# {{{ tensor product shape
+
+@dataclass(frozen=True, init=False)
```
inducer, line 571: Is the plan to replace `Hypercube` with this? (I think we should!)

modepy/shapes.py
```diff
@@ -559,4 +566,40 @@ def _submes_for_hypercube(shape: Hypercube, node_tuples):
...
+class TensorProductShape(Shape):
+    """
+    .. attribute:: bases
+
+        A :class:`tuple` of base shapes that form the tensor product.
+    """
+
+    def __init__(self, bases: Tuple[Shape, ...]) -> None:
```
inducer, line 579: * Do we genuinely allow any shapes? Triangle x Triangle? That's already 4D, so I'm kind of short on imagination in terms of what that would look like/be useful for.
* Flatten if any of the underlying things is also a `TensorProduct`?

modepy/shapes.py
```diff
@@ -559,4 +566,40 @@ def _submes_for_hypercube(shape: Hypercube, node_tuples):
...
+    def dim(self) -> int:
+        return sum(s.dim for s in self.bases)
+
+    @property
+    def nvertices(self) -> int:
+        return np.prod([s.nvertices for s in self.bases])
+
+    @property
+    def nfaces(self) -> int:
+        # FIXME: is there a formula for this?
```
inducer, line 592: For "adventurous shape" x interval, seemingly yes: One per face of the adventurous shape, plus a lid on each end. I believe this can be repeated for any number of intervals tensor-producted onto the shape.

modepy/spaces.py
```diff
@@ -88,12 +108,21 @@ def space_dim(self):
...
+@space_for_shape.register(Simplex)
+def _space_for_simplex(shape: Simplex, order: int):
+    return PN(shape.dim, order)
+
+# }}}
+
+
+# {{{ QN
+
 class QN(FunctionSpace):
```
inducer, line 126: Should this inherit from `TensorProductSpace`?

inducer (commented):
Took a quick scroll, had some questions. :)

inducer:
Also, thanks for working on this!

modepy/shapes.py
alexfikl: > Do we genuinely allow any shapes? Triangle x Triangle? That's already 4D, so I'm kind of short on imagination in terms of what that would look like/be useful for.

There's a good chance a few places will break on some random list of shapes, but I'm not sure how to restrict the choices. Maybe just limit the resulting dimension to 3?

modepy/shapes.py
alexfikl: We need a `Line` shape then, right? To use as a building block for the hypercube.

inducer:
A 1-simplex is a line/interval...

modepy/shapes.py
inducer: I'd say `Simplex(1)`.

modepy/shapes.py
inducer: How about "at most one non-line and otherwise only lines"?

modepy/__init__.py
alexfikl: Added some mention of tensor products in a782377. Also updated `examples/derivatives.py` to use the new and improved `modepy` functionality in the front page.

modepy/shapes.py
alexfikl: Modified this somewhere in 05fe4d1.

Basically the way things work with tensor products with right now has some `__new__` hackery in it:
* if the tensor product just gets one shape, it skips and just returns that shape
* for `Hypercube`, this means that in 1D it just returns `Simplex(1)`, i.e. there's no `Hypercube(1)`.
* the `TensorProductSpace` and `QN` work exactly the same.

modepy/shapes.py
alexfikl: That sounds good! Implemented something like that in b7db9ff. I think the checks are ok, but definitely needs a closer look.

modepy/shapes.py
alexfikl: Yep, implemented that in b7db9ff.

modepy/spaces.py
alexfikl: Yes! Made it that way somewhere in 05fe4d1 (that's why a lot of `register` for `QN` are gone).

alexfikl:
@inducer This should be ready for another look when you have a moment.

modepy/nodes.py
```diff
@@ -473,46 +483,75 @@ def _random_nodes_for_simplex(shape: Simplex, nnodes: int, rng=None):
...
+        for b, s in zip(space.bases, shape.bases)
+        ])
+
+
+@random_nodes_for_shape.register(TensorProductShape)
+def _random_nodes_for_tp(shape: TensorProductShape, nnodes: int, rng=None):
+    if rng is None:
+        rng = np.random.default_rng()
+
+    # FIXME: this may not be exactly uniform per dimension; any better way?
```
inducer, line 539: I'm not sure why you're tensor-producting randomly drawn nodes on each sub-shape. That'll come out in a grid, i.e. not very random. It'll also generate a different number of nodes than the user asked for.

Why not have each dimension generate the full `nnodes` and then "zip" them together?

inducer (commented):
Thanks for working on this! And sorry about the very long delay in reviewing it. This did take a healthy bit of time to review. A few comments below. This LGTM once those are addressed.

inducer:
Unsubscribing... @-mention or request review once it's ready for a look or needs attention.

modepy/nodes.py
```diff
@@ -473,46 +483,75 @@ def _random_nodes_for_simplex(shape: Simplex, nnodes: int, rng=None):
...
+    from pytools import generate_nonnegative_integer_tuples_below as gnitb
+    tuples_for_space = [node_tuples_for_space(s) for s in space.bases]
+
+    n = len(tuples_for_space)
+    r = tuple([
+        sum((tuples_for_space[n - i - 1][j] for i, j in enumerate(tp[::-1])), ())
+        for tp in gnitb([len(tp) for tp in tuples_for_space])
         ])
 
+    return r
```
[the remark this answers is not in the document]
alexfikl: I committed this for now, but completely agreed! It's been so long that it looks quite obfuscated to me too :\

I'll leave this open until something better comes to mind or we give up.

modepy/quadrature/__init__.py
```diff
@@ -205,6 +205,25 @@ def _quadrature_for_qn(space: QN, shape: Hypercube):
     assert quad.exact_to >= space.order
     return quad
 
+
+@quadrature_for_space.register(TensorProductSpace)
```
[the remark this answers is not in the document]
alexfikl: Done in cb4161b.

Had to modify `quadrature_for_pn` to use Gauss-Legendre quadrature in the 1D case because it was falling back to `XiaoGimbutasSimplexQuadrature`, which is apparently ever so slightly less accurate. Is that ok?

modepy/spaces.py
```diff
@@ -64,6 +80,71 @@ def space_for_shape(shape: Shape, order: int) -> FunctionSpace:
...
+        return (f"{type(self).__name__}("
+                f"spatial_dim={self.spatial_dim}, space_dim={self.space_dim}, "
+                f"bases={self.bases!r}"
+                ")")
+
+
+@space_for_shape.register(TensorProductShape)
+def _space_for_tensor_product_shape(
+        shape: TensorProductShape,
+        order: Union[int, Tuple[int, ...]]) -> TensorProductSpace:
```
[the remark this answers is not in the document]
alexfikl: Added a bit of documentation to 60e35cb.

Now that you mention it, I'm not convinced it's a good idea to allow this in the first place. The reason I added it was mostly so that the tests could construct different spaces through `space_for_shape` (otherwise they would need to special case tensor products).

I'm fine with removing it if you think it's cleaner.

modepy/spaces.py
```diff
@@ -126,28 +211,41 @@ class QN(FunctionSpace):
...
+                f"spatial_dim={self.spatial_dim}, order={self.order}"
+                ")")
 
 
 @space_for_shape.register(Hypercube)
-def _space_for_hypercube(shape: Hypercube, order: int):
-    return QN(shape.dim, order)
-
-# }}}
+def _space_for_hypercube(
```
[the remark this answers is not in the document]
alexfikl: Removed in 9a9ab21.

I kept it around because it would create `QN` spaces. Not sure if anyone in `meshmode` or `grudge` checks for that with `isinstance` or something.

modepy/nodes.py
```diff
@@ -473,46 +483,75 @@ def _random_nodes_for_simplex(shape: Simplex, nnodes: int, rng=None):
...
+    nnodes_per_shape += [nnodes - sum(nnodes_per_shape)]
+
+    return tensor_product_nodes([
+        random_nodes_for_shape(s, nnodes, rng=rng)
+        for s, nnodes in zip(shape.bases, nnodes_per_shape)
+        ])
 
 
 @random_nodes_for_shape.register(Hypercube)
 def _random_nodes_for_hypercube(shape: Hypercube, nnodes: int, rng=None):
```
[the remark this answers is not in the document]
alexfikl: Removed in 0911d70.

modepy/nodes.py
alexfikl: Hm, yeah, not quite sure why I thought that wouldn't work.

Cleaned up in 0911d70.

modepy/spaces.py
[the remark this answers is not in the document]
alexfikl: Yep, it broke something! Needs inducer/meshmode#290.

Should I add it back for backwards compatibility?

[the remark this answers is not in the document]
inducer: Nah, if nothing is broken, it should stay gone.

modepy/spaces.py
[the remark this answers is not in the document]
alexfikl: > How come?

From what I remember, mostly because the tuple version is specific to the tensor product and it felt weird to add that to the general interface.

It's a fairly small/self-contained/documented change though, so I don't feel particularly strongly either way.

modepy/modes.py
```diff
@@ -657,33 +658,82 @@ def grad_simplex_best_available_basis(dims, n):
...
+        #
+        # (here `dims_per_function` would be `(2, 1, 3)`), where each
+        # `self.derivatives[i]` contains the derivatives of `P_i` and the
+        # remaining functions as is (in the same order as above). Then, to
+        # get the derivatives with respect to, e.g., `(x3, x4, x5)`, we
+        # loop over the `components` of the gradient of `P_3` as
+        #
+        #   result[d + 0] = P_1 * P_2 * dP_3/dx_3 * ...
+        #   result[d + 1] = P_1 * P_2 * dP_3/dx_4 * ...
+        #   result[d + 2] = P_1 * P_2 * dP_3/dx_5 * ...
```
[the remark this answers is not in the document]
alexfikl: I tried to describe it a bit better in 
d823e67 but it should be more or less the same as what was in there before (see [here](https://github.com/inducer/modepy/blob/321b2747f86b0a2d111dd80b09a04408ede1c567/modepy/modes.py#L681-L683)).

Basically before, each `per_dim_derivatives` contained a list of
```
(f_1, ..., df_i, ..., f_n)
```
and there was not much to do besides evaluate each of them because they were all scalar. Now, `df_i` can return a tuple, with one entry per variable it depends on, so there's a bit more accounting for that.

alexfikl:
@inducer This should be ready for another look now. The tensor product basis and node stuff is still a bit convoluted, couldn't think of a nicer way to write them :disappointed:

modepy/modes.py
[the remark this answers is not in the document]
inducer: That helps, thanks!

modepy/quadrature/__init__.py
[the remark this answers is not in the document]
inducer: That's just fine. I think it might have fallen back to Grundmann-Möller, which is very general, and generally not good. :)

modepy/nodes.py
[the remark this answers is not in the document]
inducer: I think I can live with it for now.

inducer:
Thanks! Other than the `np.squeze` thing, this is ready to land from my perspective.

modepy/nodes.py
```diff
@@ -473,47 +483,68 @@ def _random_nodes_for_simplex(shape: Simplex, nnodes: int, rng=None):
...
-def _random_nodes_for_hypercube(shape: Hypercube, nnodes: int, rng=None):
+@random_nodes_for_shape.register(TensorProductShape)
+def _random_nodes_for_tp(shape: TensorProductShape, nnodes: int, rng=None):
     if rng is None:
-        rng = np.random
-    return rng.uniform(-1.0, 1.0, size=(shape.dim, nnodes))
+        rng = np.random.default_rng()
+
+    return np.stack([
+        np.squeeze(random_nodes_for_shape(s, nnodes, rng=rng))
```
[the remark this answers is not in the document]
alexfikl: Fixed and added a test in 55116b0.

inducer:
LGTM! In it goes. Thanks!

PR #42, opened 2021-08-30
Fix, test order-0 nodes

inducer:
@alexfikl Could you take a look?

### DIFF
--- a/modepy/nodes.py
+++ b/modepy/nodes.py
@@ -60,7 +60,7 @@ import numpy.linalg as la
 
 from functools import singledispatch, partial
 
-from modepy.shapes import Shape, Simplex, Hypercube
+from modepy.shapes import Shape, Simplex, Hypercube, unit_vertices_for_shape
 from modepy.spaces import FunctionSpace, PN, QN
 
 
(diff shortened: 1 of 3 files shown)


alexfikl (approved):
Looks good to me! This works automagically for the tensor product parts, right?

inducer:
> This works automagically for the tensor product parts, right?

*cough*

```
>>> mp.equispaced_nodes_for_space(mp.QN(3, 0), mp.Hypercube(3))
/home/andreas/src/modepy/modepy/nodes.py:478: RuntimeWarning: invalid value encountered in true_divide
  return (np.array(node_tuples_for_space(space), dtype=np.float64)
array([[nan],
       [nan],
       [nan]])
```

alexfikl:
> > This works automagically for the tensor product parts, right?
> 
> _cough_

Close enough!

That's because it divides the node tuples by the order [there](https://github.com/inducer/modepy/blob/93bf8c1a6986fe86dc2ec200379380a4b97bca01/modepy/nodes.py#L468-L469) to get the equidistant nodes. Maybe it's better to just do a tensor product of the 1D equidistant nodes to reuse these changes?

inducer:
There. All it takes some encouragement and a test. :P Thanks for nudging. Take another look?

alexfikl (approved):
:+1:

inducer:
Thanks for taking a look!

PR #43, opened 2021-09-25
Add more typing

alexfikl:
Most of this is cherrypicked from #41, since that was getting a bit unwieldy with unrelated changes. 

It contains mostly docs and typing changes, but it also has two functional changes:

* removes the `modepy.tools.gamma` function, since it was added in `math` from python 3.2. (mypy kept complaining about redefining it)
* fixes some uses of `__name` instead of `__name__` when raising some exceptions.

### DIFF
--- a/modepy/modes.py
+++ b/modepy/modes.py
@@ -31,13 +31,15 @@ import numpy as np
 
 from modepy.spaces import FunctionSpace, PN, QN
 from modepy.shapes import Shape, Simplex, Hypercube
-from pytools import T
 
 
 __doc__ = """This functionality provides sets of basis functions for the
 reference elements in :mod:`modepy.shapes`.
 
 .. currentmodule:: modepy
+
+.. class:: T
+
 
 Basis Retrieval
 ---------------
@@ -129,7 +131,7 @@ def _where(op_a, comp, op_b, then, else_):
 
 # {{{ jacobi polynomials
 
-def jacobi(alpha: float, beta: float, n: int, x: T) -> T:
+def jacobi(alpha: float, beta: float, n: int, x: np.ndarray) -> np.ndarray:
     r"""Evaluate `Jacobi polynomials
     <https://en.wikipedia.org/wiki/Jacobi_polynomials>`_ of type
     :math:`(\alpha, \beta)`, with :math:`\alpha, \beta > -1`, and order *n*
@@ -145,7 +147,7 @@ def jacobi(alpha: float, beta: float, n: int, x: T) -> T:
     :returns: a vector of :math:`P^{(\alpha, \beta)}_n` evaluated at all *x*.
     """
 
-    from modepy.tools import gamma
+    from math import gamma
 
     # Initial values P_0(x) and P_1(x)
     # NOTE: general formula gets a divide by 0 in the `alpha + beta == -1` case,
@@ -190,7 +192,7 @@ def jacobi(alpha: float, beta: float, n: int, x: T) -> T:
     return pl[n]
 
 
-def grad_jacobi(alpha: float, beta: float, n: int, x: T) -> T:
+def grad_jacobi(alpha: float, beta: float, n: int, x: np.ndarray) -> np.ndarray:
     """Evaluate the derivative of :func:`jacobi`, with the same meanings and
     restrictions for all arguments.
     """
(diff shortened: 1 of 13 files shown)


modepy/shapes.py
```diff
@@ -264,7 +264,8 @@ def face_normal(face: Face, normalize: bool = True) -> np.ndarray:
     volume_vertices = unit_vertices_for_shape(face.volume_shape)
     face_vertices = volume_vertices[:, face.volume_vertex_indices]
 
-    if face.dim == 0:
+    dim = getattr(face, "dim", face.volume_shape.dim - 1)
```
alexfikl, line 267: This is a bit unfortunate, but couldn't figure out how to convince mypy that `Face` is a mixin and all subclasses are expected to have a `dim` (?).

alexfikl:
Gitlab passed too https://gitlab.tiker.net/fikl2/modepy/-/pipelines/213947.

modepy/nodes.py
```diff
@@ -66,14 +67,18 @@
...
 # {{{ equidistant nodes
 
-def equidistant_nodes(dims, n, node_tuples=None):
+def equidistant_nodes(
+        dims: int, n: int,
+        node_tuples: Optional[Sequence[Tuple[int, ...]]] = None
+        ) -> np.ndarray:
     """
-    :arg dims: dimensionality of desired simplex
+    :param dims: dimensionality of desired simplex
```
inducer, line 75: :cry: ?

Not because I have a particular preference... (Is there a reason to prefer one over the other? Any documented suggestions?) but mostly because 90% of everything else uses `:arg`, and I'm reluctant to start another consistency crusade. OTOH, this one is fairly regex-able...

alexfikl: Hm, at some point I seem to remember that `arg` didn't pick up typing annotations but `param` did. It doesn't seem like that's the case anymore though (just checked).

alexfikl: Ha! Apparently [sphinx-autodoc-typehints](https://github.com/agronholm/sphinx-autodoc-typehints#how-it-works) liked `param`. The version that made it into the main repo probably removed that restriction.

alexfikl: Replaced in 5ef3e73.

inducer (commented):
Thanks! This looks great. I had one minor question, and I pushed a few things that I thought were improvements (but please do take a look). Once you tell me you like them, this is good to go.

modepy/modes.py
```diff
@@ -125,23 +127,23 @@ def _where(op_a, comp, op_b, then, else_):
 
 # {{{ jacobi polynomials
 
-def jacobi(alpha, beta, n, x):
+def jacobi(alpha: float, beta: float, n: int, x: np.ndarray) -> np.ndarray:
```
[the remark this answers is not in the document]
alexfikl: Shouldn't be too much trouble to add that, since most of the functions are multidimensional anyway.

I'll take a look.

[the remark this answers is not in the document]
inducer: Pymbolic doesn't really have type annotation yet... that's probably the tricky bit, and it's OK to call that out of scope.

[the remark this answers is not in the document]
alexfikl: Gave it a go in 35d63b2.

Not quite sure I constrained everything correctly with the `TypeVar`, so this probably requires a more careful look. Let me know what you think!

modepy/modes.py
```diff
@@ -25,15 +25,30 @@
...
+
+InputType = Union[np.ndarray, "pymbolic.primitives.Expression", float]
+InputTypeT = TypeVar("InputTypeT",
+        np.ndarray, "pymbolic.primitives.Expression", float)
+
+
 __doc__ = """This functionality provides sets of basis functions for the
 reference elements in :mod:`modepy.shapes`.
 
+.. class:: InputTypeT
```
inducer, line 46: Not sure I'm loving `InputTypeT` as a name, particularly because it occurs as output. :) `RealT`? `CoordinateT`? `BasisValueT`?

alexfikl: Renamed it to `RealValueT` in 1660f8e. Is that ok?

modepy/modes.py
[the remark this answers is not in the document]
inducer: LGTM!

modepy/modes.py
inducer: Yep, I like that. Works for me!

PR #45, opened 2021-10-26
Drop type: ignore for mypy#9975

inducer:
This'll become mergeable once mypy 0.920 is released, as that closes https://github.com/python/mypy/issues/9975.


inducer:
Depressing: It's a conservation of type-ignores here.

inducer:
The new type-ignores got dropped in #58.

PR #52, opened 2022-02-17
Update modal_decay

alexfikl:
* Update a call to the deprecated `simplex_onb_with_mode_ids`
* Added types


inducer:
Thanks! At some point, we should update this to properly make use of the shape/basis machinery, but I'll take this as a stepping stone. :)

PR #53, opened 2022-02-28
Add `reshape_array_for_tensor_product_space`

inducer:
This
- ~~flips the tensor product order (both nodes and modes), yet again.~~ This avoids a bunch of awkward reversals that were previously done.
- introduces `reshape_array_for_tensor_product_space` (and its inverse), which makes it easier to think about tensor product modes and nodes via array indices.
- It includes a test that this reshape is consistent with both mode and node ordering.
- It further includes a test that the reshape enables dimension-by-dimension operations. (shown using the Vandermonde)

Best read commit-by-commit.

cc @xywei

### DIFF
--- a/modepy/nodes.py
+++ b/modepy/nodes.py
@@ -368,24 +368,27 @@ def tensor_product_nodes(
         not only one dimensional nodes.
     """
     if isinstance(dims_or_nodes, int):
-        assert nodes_1d is not None
-        nodes: Sequence[np.ndarray] = [nodes_1d.reshape(1, -1)] * dims_or_nodes
+        if nodes_1d is None:
+            raise ValueError("nodes_1d must be supplied if the first argument "
+                                        "is the number of dimensions")
+        nodesets: Sequence[np.ndarray] = [nodes_1d.reshape(1, -1)] * dims_or_nodes
         dims = dims_or_nodes
     else:
-        assert nodes_1d is None
-        nodes = [(n.reshape(1, -1) if n.ndim == 1 else n) for n in dims_or_nodes]
-        dims = sum(n.shape[0] for n in nodes)
+        if nodes_1d is not None:
+            raise ValueError("nodes_1d must not be supplied if the first argument "
+                    "is a sequence of node arrays")
+        nodesets = [(n.reshape(1, -1) if n.ndim == 1 else n) for n in dims_or_nodes]
+        dims = sum(n.shape[0] for n in nodesets)
 
-    nnodes = len(nodes)
-    result = np.empty((dims,) + tuple([n.shape[-1] for n in nodes]))
+    result = np.empty((dims,) + tuple([n.shape[-1] for n in nodesets]))
 
     d = 0
-    for n in range(nnodes):
-        x = nodes[nnodes - 1 - n]
-        result[d:d + x.shape[0]] = x.reshape(x.shape + (1,)*n)
-        d += x.shape[0]
+    for nodes in nodesets:
+        node_dims, _ = nodes.shape
+        result[d:d+node_dims] = nodes.reshape(nodes.shape + (1,)*(dims-node_dims-d))
+        d += node_dims
 
-    return result.reshape(dims, -1)
+    return result.reshape(dims, -1, order="F").copy(order="C")
 
 
 def legendre_gauss_lobatto_tensor_product_nodes(dims: int, n: int) -> np.ndarray:
@@ -493,18 +496,18 @@ def _node_tuples_for_tp(space: TensorProductSpace):
     from pytools import generate_nonnegative_integer_tuples_below as gnitb
     tuples_for_space = [node_tuples_for_space(s) for s in space.bases]
 
-    n = len(tuples_for_space)
-
     def concat(tuples):
         return sum(tuples, ())
 
+    # ensure that these start numbering (0,0), (1,0), (i.e. x-axis first)
+    space_index_tuples = (
+            [tp[::-1] for tp in gnitb([len(tp) for tp in tuples_for_space[::-1]])])
     return tuple([
         concat((
-            tuples_for_space[n - i - 1][j]
-            for i, j in enumerate(tp[::-1])
+            tuples_for_space[i][j]
+            for i, j in enumerate(tp)
             ))
-        for tp in gnitb([len(tp) for tp in tuples_for_space])
-        ])
+        for tp in space_index_tuples])
 
 
 @equispaced_nodes_for_space.register(TensorProductSpace)
--- a/modepy/nodes.py
+++ b/modepy/nodes.py
@@ -366,6 +366,11 @@ def tensor_product_nodes(
 
         *dims_or_nodes* can contain nodes of general size ``(dims, nnodes)``,
         not only one dimensional nodes.
+
+    .. versionchanged:: 2022.1
+
+        The node ordering changed once again, and it is now accessible, via
+        :func:`modepy.tools.reshape_array_for_tensor_product_space`.
     """
     if isinstance(dims_or_nodes, int):
         if nodes_1d is None:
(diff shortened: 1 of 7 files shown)


inducer:
Hmm, I'm a bit for how to proceed here. I sort of like the version of this that includes the [node ordering flip](https://github.com/inducer/modepy/compare/flip-tensor-product-ordering?expand=1), because it removes a bunch of hard-to-think-about reversals. At the same time, that version currently bombs the meshmode tests for node/vertex numbering inconsistency:

https://github.com/inducer/modepy/runs/5419405818?check_suite_focus=true

There are two ways out of this:
- Fix meshmode to number vertices to match modepy
- Keep the reversals and fix the reshape

@alexfikl @xywei Thoughts on this?

inducer:
I guess this comes down to the question of whether we'd like to keep or ditch the reversals.

alexfikl:
> There are two ways out of this:
> 
>     * Fix meshmode to number vertices to match modepy 
>     * Keep the reversals and fix the reshape
> 
> @alexfikl @xywei Thoughts on this?

Why does `meshmode` break? I thought it asked `modepy` for nodes and vertices and node tuples, so it should stay consistent with it.

inducer:
> Why does `meshmode` break?

I actually don't know... I haven't yet dug into it and how fixable it is. Shallowly, it complains about node/vertex ordering inconsistency.

alexfikl:
> I actually don't know... I haven't yet dug into it and how fixable it is. Shallowly, it complains about node/vertex ordering inconsistency.

Yeah, I remember that happening when there was some inconsistency between how the basis and the nodes were laid out.. but switching both should have worked. Hm..

xywei:
> because it removes a bunch of hard-to-think-about reversals

What benefits do reversals provide? I guess it comes down to whether orientation matter in `_HYPERCUBE_FACES`.

inducer:
> What benefits do reversals provide? I guess it comes down to whether orientation matter in _HYPERCUBE_FACES.

Other than "the tests pass with them", I'm not sure.

inducer:
So I think I figured out what was missing to make the non-flipped tensor products (and hence the reshape code) work. The face-to-volume map also needed its "spanning vertices" flipped, here:

https://github.com/inducer/modepy/blob/541be832ce331dab699c7a156fb003a6e63c1d33/modepy/shapes.py#L481

See also the latest push:

https://github.com/inducer/modepy/compare/fe1a949570d7bb733dd26799801e4493f4565f66..541be832ce331dab699c7a156fb003a6e63c1d33

inducer:
I take all that back. We *want* those reversals, definitely. With them (i.e. on `main`), our nodes and node tuples are:
```
>>> import modepy as mp
>>> shp = mp.Hypercube(3)
>>> mp.node_tuples_for_space(mp.QN(3, 1))
((0, 0, 0), (1, 0, 0), (0, 1, 0), (1, 1, 0), (0, 0, 1), (1, 0, 1), (0, 1, 1), (1, 1, 1))
>>> mp.equispaced_nodes_for_space(mp.QN(3, 1), shp) *0.5 +0.5
array([[0., 1., 0., 1., 0., 1., 0., 1.],
       [0., 0., 1., 1., 0., 0., 1., 1.],
       [0., 0., 0., 0., 1., 1., 1., 1.]])
```
Without them (i.e. with c3357c8bbad628e85058b3b96beea086533c0ab9), our nodes are
```
>>> import modepy as mp
>>> shp = mp.Hypercube(3)
>>> mp.node_tuples_for_space(mp.QN(3, 1))
((0, 0, 0), (0, 0, 1), (0, 1, 0), (0, 1, 1), (1, 0, 0), (1, 0, 1), (1, 1, 0), (1, 1, 1))
>>> mp.equispaced_nodes_for_space(mp.QN(3, 1), shp) *0.5 +0.5
array([[0., 0., 0., 0., 1., 1., 1., 1.],
       [0., 0., 1., 1., 0., 0., 1., 1.],
       [0., 1., 0., 1., 0., 1., 0., 1.]])
```
IIRC, that's what our last direction flip was about. Why would we start numbering nodes so that a step in the Z direction comes first? Super weird, not a good idea. The node order stays as it is on main.

inducer:
That said, we do still have a bug:
```
mp.TensorProductSpace(tuple([mp.PN(1, 3+i) for i in range(dim)]))
```
generates 5 nodes along the X axis and 4 nodes along the Y axis. I would've expected the opposite.

inducer:
OK, finally. I think this makes sense. It gets rid of the reversals, but the net node ordering doesn't actually change. It fixes the issue that inhomogeneous tensor products get their axes reversed, as described above. And the tensor product reshapes serve as tests that the tensor product mode and node ordering make sense.

modepy/nodes.py
```diff
@@ -366,26 +366,30 @@ def tensor_product_nodes(
...
+
+    .. versionchanged:: 2022.1
+
+        The node ordering changed once again, and it is now accessible, via
+        :func:`modepy.tools.reshape_array_for_tensor_product_space`.
     """
     if isinstance(dims_or_nodes, int):
         assert nodes_1d is not None
-        nodes: Sequence[np.ndarray] = [nodes_1d.reshape(1, -1)] * dims_or_nodes
+        nodesets: Sequence[np.ndarray] = [nodes_1d.reshape(1, -1)] * dims_or_nodes
```
alexfikl, line 377: The first version of this PR made that assert into an exception with a nice message. Worth adding it back?

modepy/tools.py
```diff
@@ -415,4 +439,67 @@ def estimate_lebesgue_constant(n, nodes, shape=None, visualize=False):
...
+    :arg result: *ary* reshaped with axis number *axis* replaced by
+        a tuple of dimensions matching the dimensions of the spaces
+        making up the tensor product. Variation of the represented
+        function along a given dimension will be represented by variation
+        of array entries along the corresponding array axis.
+    """
+    if axis < 0:
+        axis += len(ary.shape)
+    if not (0 <= axis < len(ary.shape)):
+        raise ValueError("invalid axis specified")
```
alexfikl, line 475: Add a check that the last dimension matches `space_dim` before doing the reshape?

modepy/tools.py
```diff
@@ -415,4 +439,67 @@ def estimate_lebesgue_constant(n, nodes, shape=None, visualize=False):
...
+    return ary.reshape(
+            (ary.shape[:axis]
+                + tuple([s.space_dim for s in space.bases])
+                + ary.shape[axis+1:]),
+            order="F")
```
alexfikl, line 480: Does this need a `.copy(order="C")`? Not sure how it's being used, so maybe not.

alexfikl (approved):
Added a couple of ignorable nitpicks, but otherwise this looks good to me!

modepy/tools.py
inducer: The intended use case is to reshape group arrays (within `DOFArray`) so that tensor product operators can be applied efficiently (i.e. with way fewer flops), as tested by `test_tensor_product_vdm_dim_by_dim`. Given this use case, I don't think the extra copy would be beneficial. For one, a pyopencl `Array` couldn't even perform the operation. And even if it could, it would impose a cost in memory traffic that just isn't necessary in general.

modepy/nodes.py
inducer: https://github.com/inducer/modepy/compare/ef309cf2dd6e479823af74348b42144a75d6037a..993ee847d6096bf265d998ab89229bb6927e3a4b

modepy/tools.py
inducer: https://github.com/inducer/modepy/compare/ef309cf2dd6e479823af74348b42144a75d6037a..993ee847d6096bf265d998ab89229bb6927e3a4b

inducer:
Thanks for taking a look!

PR #71, opened 2023-10-25
Make Hypercube and TensorProductElementSpace picklable

a-alveyblanc:
Closes https://github.com/inducer/meshmode/issues/391

### DIFF
--- a/modepy/test/test_tools.py
+++ b/modepy/test/test_tools.py
@@ -600,7 +600,7 @@ def test_tensor_product_vdm_dim_by_dim(dim):
 
 @pytest.mark.parametrize("dims", [1, 2, 3])
 @pytest.mark.parametrize("shape_cls", [mp.Simplex, mp.Hypercube])
-def test_shape_pickling(dims: int, shape_cls: mp.Shape) -> None:
+def test_shape_pickling(dims, shape_cls):
     import pickle
 
     shape = shape_cls(dims)
--- a/modepy/test/test_tools.py
+++ b/modepy/test/test_tools.py
@@ -596,6 +596,36 @@ def test_tensor_product_vdm_dim_by_dim(dim):
 # }}}
 
 
+# {{{ test_tensor_product_pickle
+
+def test_tensor_product_pickling():
+    import pickle
+
+    hc = mp.Hypercube(3)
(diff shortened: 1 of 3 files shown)


alexfikl:
Awesome! Can you also add a very simple test with a `pickle.dump` and `pickle.load` to make sure we don't break this again?

a-alveyblanc:
This is good to go from my end, I think.

modepy/test/test_tools.py
```diff
@@ -596,6 +596,37 @@ def test_tensor_product_vdm_dim_by_dim(dim):
...
+
+def test_tensor_product_pickling():
+    import pickle
+
+    hc = mp.Hypercube(3)
+    tp_space = mp.space_for_shape(hc, 1)
+
+    file = open(b"hypercube-shape.obj", "wb")
+    pickle.dump(hc, file)
+    file.close()
```
alexfikl, line 609:
```suggestion
    pkl = pickle.dumps(hc)
    reloded_hc = pickle.loads(pkl)
    assert hc == reloaded_hc
```
? There's no need to dump to a file. Just a small nitpick though, so feel free to just leave it as is.

a-alveyblanc: I've made this change locally for the Hypercube, but doing the same assert check with the TensorProductSpace object throws an error. The original and reloaded objects seem to be the same. It seems the bases themselves are what is causing the assertion to fail. Is this expected?

alexfikl: Hm, that's weird, can't say I have an idea of why that's failing. Can you push your WIP version and I can take a look too?

alexfikl: Ah, figured it out!

* The `mp.Shape` is a dataclass and defines equality.
* The `mp.FunctionSpace` is just a standard class, so it's equality is defined based on `id(obj)`, so it will fail.

We can define `__eq__` and `__hash__` for `FunctionSpace` as well, but it's probably not needed. I updated the test to just check the `spatial_dim` and the `space_dim` instead of the whole object.

a-alveyblanc: I see, that makes sense! Sounds good, thanks for making that change

alexfikl (approved):
Nice! Thanks for fixing this!

PR #73, opened 2024-03-07
docs: rework to include attribute types

alexfikl:
This switches to inline attribute docs like
```
#: This is an attribute.
value: float
```
and uses `autoattribute` or just `:members:` to include them. Sphinx seems to pick this up a bit better and it shows up in the docs. For example for `Face`:

Screenshot_20240307_101148

For comparison the current docs: https://documen.tician.de/modepy/shapes.html#modepy.Face

### DIFF
--- a/modepy/spaces.py
+++ b/modepy/spaces.py
@@ -5,9 +5,20 @@ Function Spaces
 ---------------
 
 .. autoclass:: FunctionSpace
-.. autoclass:: PN
-.. autoclass:: QN
+    :members:
+
 .. autoclass:: TensorProductSpace
+    :members:
+    :show-inheritance:
+
+.. autoclass:: PN
+    :members:
+    :show-inheritance:
+
+.. autoclass:: QN
+    :members:
+    :show-inheritance:
+
 .. autofunction:: space_for_shape
 """
 
@@ -48,26 +59,24 @@ from modepy.shapes import Shape, Simplex, TensorProductShape
 class FunctionSpace(ABC):
     r"""An opaque object representing a finite-dimensional function space
     of functions :math:`\mathbb{R}^n \to \mathbb{R}`.
-
-    .. attribute:: spatial_dim
-
-        :math:`n` in the above definition, the number of spatial dimensions
-        in which the functions in the space operate.
-
-    .. attribute:: space_dim
-
-        The number of dimensions of the function space.
     """
+
+    @property
+    def order(self) -> int:
+        """Polynomial degree of the function space, if any."""
+        raise AttributeError(f"{type(self).__name__} has no attribute 'order'")
 
     @property
     @abstractmethod
     def spatial_dim(self) -> int:
-        pass
+        """The number of spatial dimensions in which the functions in the space
+        operate (:math:`n` in the above definition).
+        """
 
     @property
     @abstractmethod
     def space_dim(self) -> int:
-        pass
+        """The number of dimensions of the function space."""
 
 
 @singledispatch
@@ -92,13 +101,7 @@ def space_for_shape(
 # # {{{ generic tensor product space
 
 class TensorProductSpace(FunctionSpace):
-    """
-    .. attribute:: bases
-
-        A :class:`tuple` of the base spaces that take part in the
-        tensor product.
-
-    .. automethod:: __init__
+    """A function space defined as the tensor product of lower dimensional spaces.
 
     To recover the tensor product structure of degree-of-freedom arrays (nodal
     or modal) associated with this type of space, see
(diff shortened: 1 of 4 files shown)


modepy/spaces.py
```diff
@@ -48,26 +59,24 @@
...
-
-    .. attribute:: space_dim
-
-        The number of dimensions of the function space.
     """
 
+    @property
+    def order(self) -> int:
+        """Polynomial degree of the function space, if any."""
+        raise AttributeError(f"{type(self).__name__} has no attribute 'order'")
```
alexfikl, line 67: I think we discussed that it's not a good idea to add this "order" thing to the base `FunctionSpace` class. I added it here to silence mypy, which started complaining due to some additional return value typing.

I can remove it if you still think it's a good idea.

inducer:
Thanks for this! And specifically TIL about the ability to document attributes along with their declaration. In keeping with the Zen of Python, Sphinx appears to allow *two* not entirely obvious ways to document attributes: the style you've chosen here, and the "docstring after" style.

Do you have a sense which of those is preferred? What made you go with the `#:` style? Based on nothing in particular other than personal taste, "docstring after" strikes me as more consistent with other doc styles in Python, but :shrug: I have no idea. If one of the two styles were broadly preferred by the community, I would be happy to go along with that.

modepy/spaces.py
inducer: I'm OK with keeping it. I like the 'if any', to emphasize that this is not always a well-defined notion.

alexfikl:
> Thanks for this! And specifically TIL about the ability to document attributes along with their declaration. I agree that this is a step in the right direction. In keeping with the Zen of Python, Sphinx appears to allow _two_ not entirely obvious ways to document attributes: the style you've chosen here, and the "docstring after" style.
> 
> Do you have a sense which of those is preferred? What made you go with the `#:` style? Based on nothing in particular other than personal taste, "docstring after" strikes me as more consistent with other doc styles in Python, but 🤷 I have no idea. If one of the two styles were broadly preferred by the community, I would be happy to go along with that.

Hm, to be honest, I mostly went with the `#:` style because when I found out that works, that's the style the example used. I just tried it with
```
.. autoattribute:: dim
	
	The dimension.
```

And it works just the same as far as I can tell. I'll google around a bit to see if there's any recommendation / preference with regards to this stuff (if nothing else, this seems to be the style used in the [Sphinx docs for autoattribute](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#directive-autoattribute)). From the top of my head, I kind of like it because:
* It keeps the declaration and the docs together for easy lookup.
* Documenting attributes / properties / methods sort of follow the same rules with the docs next to the thing itself.
* This can be used with `:members:` to just quickly document everything in a class without manually listing `.. autoattribute` or `.. autoproperty` or `.. automethod`, etc.

Another cool thing that I found out works is:
```
#: Generic for int subclasses
MyT = TypeVar("MyT", bound=int)
```
will add a little `alias of TypeVar(‘MyT’, bound=int)` under the docstring, so it doesn't have to explain it. Unfortunately, the thing `TypeVar(...)` thing doesn't seem to be hyperlinked at all, but maybe they'll fix that..

alexfikl:
> I agree about keeping the docs next to the thing being documented. My question is related more to
> 
> ```
> #: doc doc doc
> stuff: int
> ```
> 
> vs
> 
> ```
> stuff: int
> """doc doc doc"""
> ```

Ah, sorry! Completely misunderstood the question! I couldn't really find much on that. 

* There are a few projects using `#:` (e.g. [flask](https://github.com/pallets/flask/blob/b90a4f1f4a370e92054b9cc9db0efcb864f87ebe/src/flask/views.py#L62) seems to, numpy also when documenting [global variables](https://github.com/numpy/numpy/blob/80a21e7771f3f6a36d5fe4581b37d3d94345e28e/numpy/typing/mypy_plugin.py#L109), [sphinx itself](https://github.com/sphinx-doc/sphinx/blob/ae51974e217cc7590e1b9c68f18af3e043027c6d/sphinx/project.py#L26)), but nothing describing a style. 
* I did a Github search on `"""` too, but nothing popped up (it's also a lot more common). One thing that did come up is [PEP 224](https://peps.python.org/pep-0224/), which looks pretty similar and wasn't appreciated (?)

So not sure what to say..

alexfikl:
Hm, scratch most of that.. There's [PEP 257](https://peps.python.org/pep-0257/) for this and it says attribute docstrings are after the attribute with a triple quote. So probably go for that? :grin:

inducer:
Ugh, what an annoying bikeshed time sink. :thinking: 

- There is this [discussion](https://discuss.python.org/t/revisiting-attribute-docstrings/36413/) on the Python discourse.
    - It includes a [statement](https://discuss.python.org/t/revisiting-attribute-docstrings/36413/14) by the current maintainer of flask on moving towards 224 away from `#:`.
- On the (IMO) nicely-well-defined-but-less-practical side, there is [PEP727](https://peps.python.org/pep-0727/).

Personally, I'm now leaning towards 224-style. You?

alexfikl:
> Ugh, what an annoying bikeshed time sink. 🤔

Quite :cry: 

> Personally, I'm now leaning towards 224-style. You?

Same! I'll make the change.

inducer:
Thanks!

PR #74, opened 2024-03-08
quadrature: add type annotations and improve docs

alexfikl:
The main contribution here is adding type annotation to the quadrature folder. It also updates the docs a bit in line with #73.

One refactor: in `clenshaw_curtis`, the constructor used to be `_make(n, kind)` and I broke them off into `_make_fejer1`, `_make_fejer2` and `_make_cc`. For context, `mypy` was complaining about overwriting some variables with different types and I got confused when moving things around, so I refactored it a bit.

### DIFF
--- a/modepy/quadrature/__init__.py
+++ b/modepy/quadrature/__init__.py
@@ -2,7 +2,11 @@
 .. currentmodule:: modepy
 
 .. autoclass:: Quadrature
+    :members:
+    :special-members: __call__
+
 .. autoclass:: ZeroDimensionalQuadrature
+    :show-inheritance:
 
 .. autofunction:: quadrature_for_space
 """
@@ -31,6 +35,7 @@ THE SOFTWARE.
 """
 
 from functools import singledispatch
+from typing import Callable, Iterable, Optional
 
 import numpy as np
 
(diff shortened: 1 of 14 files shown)


modepy/quadrature/__init__.py
```diff
@@ -46,48 +51,55 @@ class QuadratureRuleUnavailable(RuntimeError):
...
+        :arg exact_to: an optional argument denoting the summed polynomial
+            degree to which the quadrature is exact. By default, `exact_to`
+            is `None` and will *not* be set as an attribute.
+        """
 
-        An array of shape *(d, nnodes)*, where *d* is the dimension
+        self.nodes: np.ndarray = nodes
+        """An array of shape *(dim, nnodes)*, where *dim* is the dimension
         of the qudrature rule. In 1D, the shape is just *(nnodes,)*.
+        """
```
inducer, line 72: :exploding_head: Sphinx reads the constructor source?

inducer:
LGTM, thanks!

modepy/quadrature/__init__.py
alexfikl: Yes it does! :grin: I'm not sure when all this magic got in there, but it seems to work.

PR #82, opened 2024-05-30
Typing for `modepy.matrices`

inducer:
- Add `multi_vandermonde` for gradient Vandermonde, deprecate using `vandermonde` for this.
- `mass_matrix`, `inverse_mass_matrix`: request `Basis`, deprecate list of functions, add guard on orthogonality.

### DIFF
--- a/modepy/matrices.py
+++ b/modepy/matrices.py
@@ -21,9 +21,13 @@ THE SOFTWARE.
 """
 
 
+from typing import Callable, Sequence, Tuple, Union
+from warnings import warn
+
 import numpy as np
 import numpy.linalg as la
 
+from modepy.modes import Basis, BasisNotOrthonormal
 from modepy.quadrature import Quadrature
 from modepy.shapes import Face
 
@@ -32,6 +36,8 @@ __doc__ = r"""
 .. currentmodule:: modepy
 
 .. autofunction:: vandermonde
+
+.. autofunction:: multi_vandermonde
 
 Vandermonde matrices are very useful to concisely express interpolation. For
 instance, one may use the inverse :math:`V^{-1}` of a Vandermonde matrix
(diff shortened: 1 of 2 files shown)


modepy/matrices.py
```diff
@@ -83,24 +92,66 @@ def vandermonde(functions, nodes):
...
+        result[:, j] = f(nodes)
+
+    return result
+
+
+def multi_vandermonde(
+            functions: Sequence[Callable[[np.ndarray], Sequence[np.ndarray]]],
+            nodes: np.ndarray
+        ) -> Sequence[np.ndarray]:
+    """Return a (generalized) Vandermonde matrix.
```
alexfikl, line 123:
```suggestion
    """Evaluate multiple (generalized) Vandermonde matrices.
```
? Or something like that so that it's the exact same sentence as for `vandermonde`.

modepy/matrices.py
```diff
@@ -83,24 +92,66 @@ def vandermonde(functions, nodes):
...
+    for j, f in enumerate(functions):
+        result[:, j] = f(nodes)
+
+    return result
+
+
+def multi_vandermonde(
+            functions: Sequence[Callable[[np.ndarray], Sequence[np.ndarray]]],
+            nodes: np.ndarray
+        ) -> Sequence[np.ndarray]:
```
alexfikl, line 122:
```suggestion
        ) -> Tuple[np.ndarray, ...]:
```
? I'm never sure what rule to use here, but generally go for "generic inputs + specific outputs"?

modepy/matrices.py
```diff
@@ -66,7 +72,10 @@
 """
 
 
-def vandermonde(functions, nodes):
+def vandermonde(
+            functions: Sequence[Callable[[np.ndarray], np.ndarray]],
+            nodes: np.ndarray
+        ) -> np.ndarray:
```
alexfikl, line 78: Couldn't add a comment below, but this could delete the sentence in the docs below with `*functions* are allowed to return...`?

alexfikl (approved):
Left a couple of random nitpicks, but this looks good to me!

Just curious, what was the motivation to split out `multi_vandermonde`?

inducer:
> Just curious, what was the motivation to split out `multi_vandermonde`?

If we had `vandermonde` keep both "modes", then the return type is a `Union`, which means user code must `assert` which of the cases of the `Union` it falls into. Pointless and messy. Two functions seem cleaner, and the control flow in the function is easier to understand.

modepy/matrices.py
inducer: Done.

modepy/matrices.py
inducer: Done.

modepy/matrices.py
inducer: Done.

modepy/matrices.py
```diff
@@ -83,24 +92,66 @@ def vandermonde(functions, nodes):
...
+        warn("Calling vandermonde on tuple-returning functions is deprecated. "
+             "This will stop working in 2025. "
+             "Call multi_vandermonde instead.",
+             DeprecationWarning, stacklevel=2)
+        return multi_vandermonde(functions, nodes)
+
+    result = np.empty((nnodes, nfunctions), f_first_values.dtype)
+    result[:, 0] = f_first_values
+
+    for j, f in enumerate(functions):
```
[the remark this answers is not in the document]
inducer: Done.

inducer:
Thanks for taking a look!

PR #83, opened 2024-06-04
Mass matrix quad

inducer:
This adds two new mass matrix functions, plus a slew of more-or-less related cleanups. Best read commit-by-commit.

### DIFF
--- a/modepy/matrices.py
+++ b/modepy/matrices.py
@@ -21,7 +21,7 @@ THE SOFTWARE.
 """
 
 
-from typing import Callable, Sequence, Tuple, Union
+from typing import Callable, Optional, Sequence, Tuple, Union
 from warnings import warn
 
 import numpy as np
@@ -151,7 +151,12 @@ def multi_vandermonde(
     return result
 
 
-def resampling_matrix(basis, new_nodes, old_nodes, least_squares_ok=False):
+def resampling_matrix(
+            basis: Sequence[Callable[[np.ndarray], np.ndarray]],
+            new_nodes: np.ndarray,
+            old_nodes: np.ndarray,
+            least_squares_ok: bool = False
+        ) -> np.ndarray:
     """Return a matrix that maps nodal values on *old_nodes* onto nodal
     values on *new_nodes*.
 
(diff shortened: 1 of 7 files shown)


modepy/matrices.py
```diff
@@ -229,17 +241,31 @@ def differentiation_matrices(basis, grad_basis, nodes, from_nodes=None):
...
-    if isinstance(grad_vdms, tuple):
-        return tuple(
-                np.asarray(la.solve(vdm.T, gv.T).T, order="C")
-                for gv in grad_vdms)
-    else:
-        return np.asarray(
-                la.solve(vdm.T, grad_vdms.T).T,
-                order="C")
+    if len(basis) != from_nodes.shape[-1]:
+        raise ValueError("basis is not unisolvent, cannot interpolate")
```
alexfikl, line 245: Put this check before the Vandermonde matrices are computed?

modepy/matrices.py
```diff
@@ -313,8 +339,68 @@ def mass_matrix(
...
+        ) -> np.ndarray:
+    r"""Using the *quadrature*, provide a matrix :math:`M` that
+    satisfies:
+
+    .. math::
+
+        \displaystyle (M \boldsymbol u)_i = \sum_j w_j \phi_i(r_j) u_j,
+
+    where :math:`\phi_i` are the (volume) Lagrange basis functions obtained from
+    *test_functions* at *nodes*, :math:`w_i` and :math:`r_i` are the
```
alexfikl, line 379:
```suggestion
    *test_functions* at *nodes*, :math:`w_j` and :math:`r_j` are the
```
? To match the indices in the formula.

modepy/matrices.py
```diff
@@ -369,34 +459,62 @@ def nodal_mass_matrix_for_face(face: Face, face_quad: Quadrature,
...
     satisfies:
 
     .. math::
 
         \displaystyle (M_f \boldsymbol u)_i = \sum_j w_j \phi_i(r_j) u_j,
 
-    where :math:`\phi_i` are the (volume) Lagrange basis functions obtained from
-    *test_functions* at *volume_nodes*, and :math:`w_i` and :math:`r_i` are the
-    weights and nodes from *face_quad*.
+    where :math:`\phi_i` are the *test_functions*, :math:`w_i` and :math:`r_i` are
```
alexfikl, line 476:
```suggestion
    where :math:`\phi_i` are the *test_functions*, :math:`w_j` and :math:`r_j` are
```
?

alexfikl (approved):
Left a couple of very nitpicky comments, but otherwise this looks nice and straightforward to me :grin:

inducer:
Thanks for taking a look!

> Also, regarding that lumping in grudge. Is it possible with this interface? If I understand correctly, it would just mean passing a Gauss-Legendre quadrature or a Lobatto-G-L to integrate in nodal/modal_quad_mass_matrix?

I'm planning to add a separate function that returns just the diagonal, after appropriate checking. But this seemed like a nice, self-contained set of changes, so I figured I'd do that separately. I'm also still puzzling over the interface.

modepy/matrices.py
inducer: Done

modepy/matrices.py
inducer: Done

PR #87, opened 2024-06-28
Migrate to pyproject and ruff

alexfikl:
This isn't done, but it starts updating `modepy` (since it's nice and small) to the new shiny linting :grin:

The code in `contrib/pov-nodes` was a bit unhappy, so I ran `ruff format` on it too.


alexfikl:
@inducer This seems to work now. It's passing on Gitlab as well: https://gitlab.tiker.net/fikl2/modepy/-/pipelines/561724

inducer:
Thx!

alexfikl:
@inducer Not sure what that pending `flake8` job is about, but it seems to be blocking auto-merge? :-s

inducer:
Ah! Thanks for pointing that out. That was in the Github settings for the branch. Gone now, replaced by ruff.

PR #96, opened 2024-09-17
Introduce `quadrature.inf` to denote infinite accuracy

inducer:
The second commit is noisy but uncontroversial. I'm mostly looking for feedback on the first.

cc @a-alveyblanc


inducer:
Surprising! It seems like nothing in meshmode or grudge (or the tests) seems to mind this.

alexfikl (approved):
The first commit looks good to me!

Just wondering about the motivation. What's the reason not to just use `float("inf")`? Might suggest adding a little doc blurb to `_Inf` to justify its existence :grin:

inducer:
> Just wondering about the motivation. What's the reason not to just use `float("inf")`? Might suggest adding a little doc blurb to `_Inf` to justify its existence 😁

Mostly to avoid introducing floats there as a valid data type, following the mantra "make invalid states unrepresentable". If we did, someone could claim their quadrature is exact to polynomial degree $3.5$. :thinking: 

I'll add a note.

alexfikl:
> Mostly to avoid introducing floats there as a valid data type, following the mantra "make invalid states unrepresentable". If we did, someone could claim their quadrature is exact to polynomial degree 3.5 . 🤔

I was thinking of something like `Literal[float("inf")]`, but apparently that's now allowed :(
https://github.com/python/typing/issues/1160

PR #100, opened 2024-10-11
Support reshaping 1d tensor products

alexfikl:
Any reason this wasn't allowed? It mostly broke when trying to access `space.bases`, but a 1D tensor product space was just a `PN` on a `Simplex(1)`.

Not sure if just returning the array is sufficient, but the tests seem to pass.


inducer:
That seems entirely fine to me. Thanks for putting it together!

PR #102, opened 2024-12-02
Add `Basis.derivatives`

inducer:
cc @a-alveyblanc

### DIFF
--- a/modepy/matrices.py
+++ b/modepy/matrices.py
@@ -131,6 +131,11 @@ def multi_vandermonde(
     A sequence of matrices is returned--i.e. this function
     works directly on :func:`modepy.Basis.gradients` and returns
     a tuple of matrices.
+
+    .. note::
+
+        If only one of the matrices is needed, it may be convenient to instead call
+        :func:`vandermonde` with the result of :meth:`Basis.derivatives`.
     """
 
     nnodes = nodes.shape[-1]
--- a/modepy/modes.py
+++ b/modepy/modes.py
@@ -32,6 +32,7 @@ from functools import partial, singledispatch
 from typing import (
     TYPE_CHECKING,
     TypeVar,
+    cast,
 )
 
 import numpy as np
@@ -561,10 +562,10 @@ class _TensorProductBasisFunction:
         assert len(self.dims_per_function) == len(self.functions)
 
     @property
-    def ndim(self):
+    def ndim(self) -> int:
         return sum(self.dims_per_function)
 
-    def __call__(self, x):
+    def __call__(self, x: np.ndarray) -> np.ndarray:
         assert x.shape[0] == self.ndim
 
         n = 0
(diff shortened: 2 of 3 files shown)


inducer:
It also simplifies the tensor product derivative spaghetti by quite a bit.

modepy/test/test_modes.py
```diff
@@ -198,6 +199,13 @@ def test_basis_grad(dim, shape_cls, order, basis_getter):
             logger.info("error: %.5e", err)
             eoc_rec.add_data_point(h, err)
 
+            for idiff_axis in range(shape.dim):
+                dr_bf = basis.functions_dr(idiff_axis)[ifunc](r)
+                ref = gradbf_v[idiff_axis]
+                err = la.norm(dr_bf - ref, 2)
+                if la.norm(ref, 2) > 1e-13:
+                    err = err/la.norm(ref, 2)
```
inducer, line 207: Whoops. This isn't testing anything. :facepalm: On it.

inducer: Now actually tested

modepy/modes.py
```diff
@@ -766,6 +791,28 @@ def gradients(self) -> tuple[BasisGradientType, ...]:
         Each function returns a tuple of derivatives, one per reference axis.
         """
 
+    def functions_dr(self, axis: int) -> tuple[BasisFunctionType, ...]:
```
inducer, line 794: Still unsure about the name. Suggestions?

inducer: Maybe just `derivatives`? `deriv_functions`?

inducer: I agree, that's probably the sanest. Renamed to that in the latest.

modepy/matrices.py
```diff
@@ -131,6 +131,12 @@ def multi_vandermonde(
     A sequence of matrices is returned--i.e. this function
     works directly on :func:`modepy.Basis.gradients` and returns
     a tuple of matrices.
+
+    .. note::
+
+        Depending on the use case, it may be convenient to instead call
+        :func:`vandermonde` with the result of :meth:`Basis.derivatives`,
+        which is mathematically equivalent.
```
alexfikl, line 139: Maybe phrase this more like "it also works with Basis.derivatives"? It's a bit unclear what use cases could / should work.

alexfikl (approved):
Left a little documentation nitpick, but otherwise this looks good to me!

modepy/modes.py
alexfikl: `axis_derivative_functions`? 

I'm also leaning towards `derivatives`, so that's perfectly fine by me!

inducer: `axis_derivatives` is another option, but it's unlikely that we'd have any other type (directional?) derivative, so :shrug:. I say let's leave it as `derivatives`.

modepy/matrices.py
inducer: Did some rephrasing, though I'm not sure I covered what you asked for here. This is in the docstring for `multi_vandermonde`, which won't work with `Basis.derivatives`. I did specify the use case more clearly. LMK if this works.

alexfikl: Yeah, that's a lot clearer to me. Thanks!

PR #103, opened 2024-12-05
Add generic nodal/modal bilinear form routines

a-alveyblanc:
It would be helpful to have a more general version of `nodal_quad_mass_matrix`. My specific use case for this routine is generating operators representing bilinear forms that have derivatives on the test functions but not necessarily on the trial functions. More generally, this can be useful to generate operators representing bilinear forms whose test and trial functions are not the same.

Functionality of `nodal_quad_mass_matrix` and `modal_quad_mass_matrix` are updated to use the new, general routines.


a-alveyblanc:
Not sure why ruff was failing here. It wasn't failing locally, and I didn't change any of the files the ruff CI test was complaining about. Fixed them anyway, but happy to revert if it seems like a CI bug.

The changes in `contrib/weights-from-festa-sommariva` were the ones needed to make ruff happy.

alexfikl:
> Not sure why ruff was failing here. It wasn't failing locally, and I didn't change any of the files the ruff CI test was complaining about. Fixed them anyway, but happy to revert if it seems like a CI bug.
> 
> The changes in `contrib/weights-from-festa-sommariva` were the ones needed to make ruff happy.

That's a new check in ruff 0.8.2: https://github.com/astral-sh/ruff/pull/14611. Thanks for fixing it!

modepy/test/test_matrices.py
```diff
@@ -118,6 +118,70 @@ def test_tensor_product_diag_mass_matrix(shape: mp.Shape) -> None:
...
+            )
+        else:
+            nodes = mp.legendre_gauss_tensor_product_nodes(shape.dim, order)
+    elif isinstance(shape, mp.Simplex):
+        if nodes_on_bdry:
+            nodes = mp.warp_and_blend_nodes(shape.dim, order)
+        else:
+            nodes = mp.VioreanuRokhlinSimplexQuadrature(order, shape.dim).nodes
+    else:
+        raise AssertionError()
```
inducer, line 152: `mp.edge_clustered_nodes_for_space`?

inducer (commented):
Thanks, this looks great! Just a few style quibbles below.

a-alveyblanc:
Seems like `mypy` also now has issue with something I didn't touch. I'll take a closer look later today.

modepy/test/test_matrices.py
a-alveyblanc: Updated. Though, for the case where `nodes_on_bdry` is false, we still have to do some checking unless I'm overlooking something.

alexfikl:
> Seems like `mypy` also now has issue with something I didn't touch. I'll take a closer look later today.

Those seem to be due to an issue with some new numpy typing changes: https://github.com/numpy/numpy/issues/27957.

modepy/matrices.py
```diff
@@ -351,6 +354,77 @@ def mass_matrix(
...
+
+    where :math:`\phi_i` are the Lagrange basis functions obtained from
+    *test_functions* at *nodes*, :math:`w_j` and :math:`r_j` are the weights
+    and nodes from *quadrature*, and :math:`u_j` are point values of the trial
+    function at the *quadrature* nodes.
+    """
+    if len(test_functions) != nodes.shape[1]:
+        raise ValueError("volume_nodes not unisolvent with trial_functions")
+
+    vdm_out = vandermonde(trial_functions, nodes)
```
inducer, line 399: I think something is wonky here. The docstring claims that this is the Lagrange test basis, but this uses the trial functions. I think part of this is because the test functions may be the derivative, but the Vandermonde of interest should come from the, um, non-derivatives. Maybe what you're looking for here is an extra argument to supply the basis for the computation of this Vandermonde?

modepy/matrices.py
```diff
@@ -351,6 +354,77 @@ def mass_matrix(
     return la.inv(inverse_mass_matrix(basis, nodes))
 
 
+def modal_quad_bilinear_form(
```
inducer, line 357: I'm thinking about the naming of this. I'm not sure "bilinear form" is that great a fit, because the trial functions don't appear. I thought of "projection", but that also misses the mark, because the test functions could be derivatives. In a word, I'm not sure. I can probably be convinced to go along with "bilinear form", but I figured it might be good to at least discuss the naming.

(@alexfikl, if you're reading along, please feel free to chime in too)

a-alveyblanc: I'm also not the biggest fan of the name, but I couldn't come up with anything better. `modal_quadrature_operator` maybe?

a-alveyblanc:
The first go was a confusing amalgamation of two versions of applying an operator to evaluate a bilinear form. They should now be cleanly separated with names that do a better job of describing what's happening.

What I mean by "two versions" is:
1. We can compute a "half" operator, i.e. the test functions evaluated at quadrature nodes times quadrature weights, and apply that to a trial solution evaluated at quadrature nodes (implemented as `modal_quadrature_operator` and `nodal_quadrature_operator`), or
2. We can apply a "full" operator, i.e. compute the inner product of test and trial functions using quadrature, and apply that to the coefficients of the trial solution at discretization nodes (implemented as `modal_quadrature_bilinear_form` and `nodal_quadrature_bilinear_form`)

Hopefully this is more clear. Happy to convert to a draft if we think more work needs to be done

modepy/matrices.py
```diff
@@ -367,12 +482,7 @@ def modal_quad_mass_matrix(
 
     .. versionadded :: 2024.2
     """
-    modal_mass_matrix = np.empty((len(test_functions), len(quadrature.weights)))
-
-    for i, test_f in enumerate(test_functions):
-        modal_mass_matrix[i] = test_f(quadrature.nodes) * quadrature.weights
-
-    return modal_mass_matrix
+    return modal_quadrature_operator(quadrature, test_functions)
```
alexfikl, line 485: If this `modal_quadrature_operator` is the same as the `modal_quad_mass_matrix`, why introduce a new function? (Same for `nodal_quad_mass_matrix`)

modepy/matrices.py
```diff
@@ -351,6 +355,117 @@ def mass_matrix(
...
+
+    .. math::
+
+        \displaystyle (A \boldsymbol u)_i = \sum_k,j (w_k \psi_i(r_k)
+        \phi_j(r_k)) u_j,
+
+    where :math:`\psi_i` are from *test_functions*, :math:`\phi_j` are from
+    *trial_functions*, :math:`r_k` and :math:`w_k` are nodes and weights
+    from *quadrature*, and :math:`u_j` are modal coefficients of a trial
+    solution.
```
alexfikl, line 429: Maybe write this in a "bilinear form" format, i.e. $`(u, v)_A = \sum u_i A_{ij} v_j`$, and then the $A$ matrix explicitly?

alexfikl (commented):
Just took a quick look and left some comments :grin:

modepy/matrices.py
alexfikl: Where is this `modal_quad_bilinear_form` supposed to be used? I'm guessing the `nodal_quad_bilinear_form` is used to define a proper bilinear form somewhere?

a-alveyblanc: `modal_quad_bilinear_form` doesn't have any uses right now other than returning the operator that's transformed into a nodal operator.

`nodal_quad_bilinear_form` will be used in a grudge update that evaluates operators using quadrature (https://github.com/inducer/grudge/pull/362 is the staging area for this, although a little outdated at this point). While `nodal_quadrature_operator` could be used for everything, `nodal_quad_bilinear_form` makes the corresponding grudge code cleaner when there is no overintegration.

TLDR; `modal_quad_bilinear_form`'s current use is in `nodal_quad_bilinear_form`. `nodal_quad_bilinear_form` is used to evaluate DG operators in grudge when there is no overintegration. `nodal_quadrature_operator` is used when there is overintegration.

Also, the original comment is outdated now since `modal_quad_bilinear_form` *does* use trial functions (as of [c063d23](https://github.com/inducer/modepy/pull/103/commits/c063d234325d04823773c3ea614ee7f2e0bb8586)).

modepy/matrices.py
a-alveyblanc: Mostly because of the name. As a user, `*_quad_mass_matrix` would make me assume this is only used for generating an operator whose action is equivalent to applying a mass matrix. 

I suppose I can set `*_quad_mass_matrix = *_quadrature_operator` and get rid of `*_quad_mass_matrix` functions?

alexfikl: I don't know, I'm personally not very bothered by the existing naming, especially since it gives a pretty clear idea of some use for the function (i.e. everybody knows what a mass matrix is), while "quadrature_operator" is very generic sounding.

> I suppose I can set *_quad_mass_matrix = *_quadrature_operator and get rid of *_quad_mass_matrix functions?

If these go away, the nice way to do it is to add a `warn("This function is going away on {deadline}. Use {other_function} instead.")` sort of thing so that users have time to migrate. :grin:

a-alveyblanc: Sure, I agree. I guess my main point is that `*_quadrature_operator` can be used to generate a stiffness matrix and a mass matrix while `*_quad_mass_matrix` is restricted to only generating a mass matrix (since it requires the vandermonde built with `test_functions` is invertible).

I agree that the naming is pretty generic. I'm open to suggestions if you've got any! :)

alexfikl: Also, if you do decide to deprecate the `*_quad_mass_matrix`, I would suggest using `*_quad_operator` for the new names, since most of the functions in here seem to use the short name?

modepy/matrices.py
```diff
@@ -351,6 +355,117 @@ def mass_matrix(
     return la.inv(inverse_mass_matrix(basis, nodes))
 
 
+def modal_quadrature_operator(
```
inducer: - `test_func_integration_matrix`?
- `modal_test_matrix_quad`?

modepy/matrices.py
```diff
@@ -351,6 +355,117 @@ def mass_matrix(
...
+    where :math:`\phi_i` are the *test_functions* at the nodes
+    :math:`r_j` of *quadrature*, with corresponding weights :math:`w_j`, and
+    :math:`u_j` is a trial solution evaluated at the *quadrature* nodes.
+    """
+    modal_operator = np.empty((len(test_functions), len(quadrature.weights)))
+
+    for i, test_f in enumerate(test_functions):
+        modal_operator[i] = test_f(quadrature.nodes) * quadrature.weights
+
+    return modal_operator
```
inducer, line 378: `return np.array([...])`?

modepy/matrices.py
```diff
@@ -351,11 +355,131 @@ def mass_matrix(
...
+
+
+def nodal_quad_operator(
```
inducer, line 381: Have something like this instead?
```
def modal_to_nodal(basis, nodes, matrix):
    vdm = vandermonde(proj_functions, nodes)

    return la.solve(
        vdm.T, matrix
    )
```
and get rid of the `nodal_...` versions?

modepy/matrices.py
```diff
@@ -351,11 +355,131 @@ def mass_matrix(
...
+    *quadrature*. The matrix :math:`A` satisfies
+
+    .. math::
+
+        \displaystyle (u, \psi_i)_A = \sum_{j} A_{ij} u_j,
+
+    where :math:`u_i` are nodal coefficients of a trial solution.
+    """
+    if len(test_functions) != nodes.shape[1]:
+        raise ValueError("volume_nodes not unisolvent with proj_functions")
```
inducer, line 468:
```suggestion
    if len(proj_functions) != nodes.shape[1]:
        raise ValueError("volume_nodes not unisolvent with proj_functions")
```
(but this is probably going away anyway)

modepy/matrices.py
```diff
@@ -351,11 +355,131 @@ def mass_matrix(
...
+
+    modal_operator = modal_quad_bilinear_form(
+        quadrature, test_functions, trial_functions)
+
+    vdm = vandermonde(proj_functions, nodes)
+
+    return la.solve(vdm.T, modal_operator) @ la.inv(vdm)
+
+
 def modal_quad_mass_matrix(
```
inducer, line 478: I'd vote in favor of deprecating the `_mass_matrix` functions.

modepy/matrices.py
inducer, line 1: While we're at it, deprecate the `*face_mass*` things, too, and (maybe) have a basis-with-node-transform-composer?

inducer:
Unsubscribing... @-mention or request review once it's ready for a look or needs attention.

a-alveyblanc:
@inducer This is ready for another look

modepy/matrices.py
```diff
@@ -78,8 +76,11 @@
 """
 
 
+BasisFunctionType = Callable[[np.ndarray], np.ndarray]
```
inducer, line 79:
```suggestion
NodalFunction: TypeAlias = Callable[[np.ndarray], np.ndarray]
```
?
So two things:

- make it a type alias
- rename

inducer, line 79: Move to `modes`, use there, too?

modepy/matrices.py
```diff
@@ -351,61 +352,122 @@ def mass_matrix(
...
-            test_functions: Sequence[Callable[[np.ndarray], np.ndarray]],
-        ) -> np.ndarray:
-    r"""Using the *quadrature*, provide a matrix :math:`M` that
-    satisfies:
+def nodal_quadrature_matrix(
+        quadrature: Quadrature,
+        test_basis: Basis,
+        test_derivative_ax: int | None = None,
+        nodes: np.ndarray | None = None,
+        mapping_function: Callable[[np.ndarray], np.ndarray] | None = None
```
inducer, line 360: Can you explain what the story is with the `mapping_function`? (It should also be documented.) To be mathematically valid, the mapping determinant be needed? I'm not sure I'm understanding the use case.

modepy/matrices.py
```diff
@@ -351,61 +352,122 @@ def mass_matrix(
     return la.inv(inverse_mass_matrix(basis, nodes))
 
 
-def modal_quad_mass_matrix(
```
inducer, line 354: I don't recall what we had agreed upon regarding the availability of this "modal" business. Could you remind me?

modepy/matrices.py
```diff
@@ -351,61 +352,122 @@ def mass_matrix(
...
-            quadrature: Quadrature,
-            test_functions: Sequence[Callable[[np.ndarray], np.ndarray]],
-        ) -> np.ndarray:
-    r"""Using the *quadrature*, provide a matrix :math:`M` that
-    satisfies:
+def nodal_quadrature_matrix(
```
inducer, line 355:
```suggestion
def nodal_quadrature_test_matrix(
```
?

modepy/matrices.py
```diff
@@ -351,61 +352,122 @@ def mass_matrix(
...
-    .. versionadded :: 2024.2
+        \displaystyle (u, \psi_i)_A = \sum_{j} A_{ij} u_j,
+
+    where :math:`u_i` are nodal coefficients.
+
+    An optional *mapping_function* can be supplied to map the *quadrature* nodes
+    evaluated by *test_basis* before constructing the matrix.
+
+    An optional set of *output_nodes* can be supplied if the result should live
+    on a domain different from the domain of *input_nodes*.
```
inducer, line 430: Too colloquial.

modepy/matrices.py
```diff
@@ -351,61 +352,122 @@ def mass_matrix(
...
+
+    Test and trial function derivatives can be requested by supplying either
+    *test_derivative_ax* or *trial_derivative_ax* as an integer corresponding to
+    an enumerated spatial derivative ax. For example, *test_derivative_ax = 0*
+    would use *test_basis* differentiated with respect to :math:`x` while
+    *trial_derivative_ax = 2* would use *trial_basis* differentiated with
+    respect to :math:`z`.
```
inducer, line 437: Not sure this is worth it, vs. having the user just pass in the derivative function set.

modepy/test/test_matrices.py
```diff
@@ -33,49 +33,6 @@
...
-            nodes = mp.warp_and_blend_nodes(shape.dim, order)
-        else:
-            nodes = mp.VioreanuRokhlinSimplexQuadrature(order, shape.dim).nodes
-
-    else:
-        raise AssertionError()
-
-    basis = mp.orthonormal_basis_for_space(space, shape)
-
-    quad_mass_mat = mp.nodal_quad_mass_matrix(quad, basis.functions, nodes)
```
inducer, line 70: These are deprecated, but I'm not sure we should delete their tests just yet?

modepy/matrices.py
```diff
@@ -429,25 +491,51 @@ def spectral_diag_nodal_mass_matrix(
     return quadrature.weights
 
 
-def modal_mass_matrix_for_face(
-            face: Face, face_quad: Quadrature,
-            trial_functions: Sequence[Callable[[np.ndarray], np.ndarray]],
-            test_functions: Sequence[Callable[[np.ndarray], np.ndarray]]
+# {{{ deprecated remove in 2025-ish
```
inducer, line 494:
```suggestion
# {{{ deprecated remove in 2026-ish
```

inducer (commented):
Some thoughts from a quick scroll below.

modepy/matrices.py
a-alveyblanc: IIRC, we decided to absorb the modal routines into the nodal routines since there wasn't a use-case outside of the nodal routines.

modepy/matrices.py
a-alveyblanc: For a specific example, `mapping_function` is a mapping from face -> volume, obtained from face.map_to_volume. This (or accepting the mapped nodes) is necessary to get rid of the `*_mass_matrix_for_face` routines.

modepy/matrices.py
a-alveyblanc: Looks like it's already in `modes`: https://github.com/inducer/modepy/blob/635dc594f6a862c7eab98a4a793a7de5e30f6617/modepy/modes.py#L748-L749

modepy/test/test_matrices.py
a-alveyblanc: Whoops. This was a mistake, thanks for catching it!

modepy/matrices.py
```diff
@@ -351,61 +353,115 @@ def mass_matrix(
...
 
+    vdm = vandermonde(test_basis_functions, nodes)
 
-def nodal_quad_mass_matrix(
+    mapped_nodes = (
+        mapping_function(quadrature.nodes)
+        if mapping_function is not None else quadrature.nodes
+    )
+
+    modal_operator = np.array([
```
inducer, line 392:
```suggestion
    modal_operator = vandermonde().T * weights
```

inducer:
Thanks, LGTM!

PR #104, opened 2024-12-12
fix: new mypy errors due to numpy typing

alexfikl:
I was looking into how to fix the mypy errors from #103. It seems like the issue it's hitting is due to redefinitions, .e.g
```
x = np.ones(5)
x = x.reshape(-1, 1)
```
is now a type error because of the different shapes of the arrays. I mostly "fixed" this by forcing the shape to be `tuple[int, ...]` in all cases, but I imagine there's a nicer way to do it..

@inducer Maybe worth just cherrypicking this into #103 if the mypy test pass, since the ruff tests may still fail.


inducer:
Since #103 has another round of review going, I figure we can get this in first. I don't love the numpy typing fixes, but I'm not sure I have a particularly better idea.

PR #105, opened 2024-12-13
Fix reshaping/unreshaping for tensor product spaces in 1D

a-alveyblanc:
`unreshape_array_for_tensor_product_space` currently fails for 1D. It checks `space.bases`, but in 1D this is `PN`, which doesn't have a `bases` attribute. This leads to a bunch of unnecessary checks on spatial dimension before attempting to reshape in an upstream package to guarantee we don't fail.

`reshape_array_for_tensor_product_space` works in the 1D case, but does some unnecessary work since the input array and returned array will be the same.

If the spatial dimension of the space is 1, we should just return the array in both cases since there's nothing to do. This PR updates corresponding functions to do just that. Also updates one of the tests to test for `unreshape_array_for_tensor_product_space` to test the 1D case.

### DIFF
--- a/modepy/tools.py
+++ b/modepy/tools.py
@@ -491,6 +491,8 @@ def reshape_array_for_tensor_product_space(
         function along a given dimension will be represented by variation
         of array entries along the corresponding array axis.
     """
+    if space.spatial_dim == 1:
+        return ary
 
     ndim = len(ary.shape)
     if axis < 0:
@@ -518,6 +520,8 @@ def unreshape_array_for_tensor_product_space(
     """Undoes the effect of :func:`reshape_array_for_tensor_product_space`,
     given the same *space* and *axis*.
     """
+    if space.spatial_dim == 1:
+        return ary
 
     n_tp_axes = len(space.bases)
     naxes = len(ary.shape) - n_tp_axes + 1
--- a/modepy/tools.py
+++ b/modepy/tools.py
@@ -70,7 +70,7 @@ import numpy.linalg as la
 from pytools import MovedFunctionDeprecationWrapper, memoize_method
 
 import modepy.shapes as shp
-from modepy.spaces import TensorProductSpace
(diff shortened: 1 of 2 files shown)


modepy/tools.py
```diff
@@ -491,6 +491,8 @@ def reshape_array_for_tensor_product_space(
         function along a given dimension will be represented by variation
         of array entries along the corresponding array axis.
     """
+    if space.spatial_dim == 1:
```
inducer, line 494:
```suggestion
    if len(space.bases) == 1:
```
This handles e.g. a trivial tensor product of triangle (2D).

modepy/tools.py
```diff
@@ -518,6 +520,8 @@ def unreshape_array_for_tensor_product_space(
     """Undoes the effect of :func:`reshape_array_for_tensor_product_space`,
     given the same *space* and *axis*.
     """
+    if space.spatial_dim == 1:
```
inducer, line 523:
```suggestion
    if len(space.bases) == 1:
```

modepy/tools.py
a-alveyblanc: This would fail for the above reason, i.e. in the 1D case `PN` does not have a `bases` attribute.

modepy/tools.py
a-alveyblanc: See other comment.

modepy/tools.py
a-alveyblanc: If we drop support for being able to create 1D tensor-product element groups and discretizations in upstream packages, then we won't need this change.

inducer: Sorry, I clearly had not read the PR description. In that case, the first thing that needs fixing is the type annotation of this function. And next, the proper spelling of this test is `if not isinstance(space, TensorProductSpace)`.

Analogous below, I think.

inducer:
LGTM, thanks!

PR #107, opened 2025-01-28
Add Vioreanu-Rokhlin quadrature finding

### LINKED ISSUE #106: Implement quadrature finding based on Vioreanu-Rokhlin
[Reference paper](https://doi.org/10.1137/110860082)

This will require a few pieces:

- Something to compute accurate integrals of the desired function basis.
- Something to set up the eigenvalue problem for the multiplication.
- An implementation of Gauss-Newton to iterate towards a better rule.

I think ideally this should be factored so that all the pieces are individually usable.

cc @a-alveyblanc @shawnL00

inducer:
To do:

- [x] Get it to work
- [x] Add a real test for the node finder (Lebesgue constant maybe?)
- [x] ~~Improve basis orthogonalization~~ (@inducer thinks this can't be improved while still representing functions as linear combinations of monomials. Using PKDOs on a simplex avoids this. Unverified conjecture: an orthonormal basis on a mismatched domain will still make this better, e.g. tensor product Chebyshev on a simplex.)
- [x] Add Gauss-Newton finder
- [x] Add proper documentation

Closes #106.

cc @ShawnL00 @a-alveyblanc

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -4,9 +4,9 @@ modepy: Basis Functions, Node Sets, Quadratures
 .. image:: https://gitlab.tiker.net/inducer/modepy/badges/main/pipeline.svg
     :alt: Gitlab Build Status
     :target: https://gitlab.tiker.net/inducer/modepy/commits/main
-.. image:: https://github.com/inducer/modepy/workflows/CI/badge.svg?branch=main&event=push
+.. image:: https://github.com/inducer/modepy/actions/workflows/ci.yml/badge.svg
     :alt: Github Build Status
-    :target: https://github.com/inducer/modepy/actions?query=branch%3Amain+workflow%3ACI+event%3Apush
+    :target: https://github.com/inducer/modepy/actions/workflows/ci.yml
 .. image:: https://badge.fury.io/py/modepy.png
     :alt: Python Package Index Release Page
     :target: https://pypi.org/project/modepy/
--- a/modepy/quadrature/construction.py
+++ b/modepy/quadrature/construction.py
@@ -1,0 +1,451 @@
+"""
+.. versionadded:: 2025.2
+
+.. autodata:: Integrand
+.. autoclass:: LinearCombinationIntegrand
+.. autofunction:: linearly_combine
+.. autofunction:: orthogonalize_basis
+.. autofunction:: adapt_2d_integrands_to_complex_arg
+.. autofunction:: guess_nodes_vioreanu_rokhlin
+.. autofunction:: find_weights_undetermined_coefficients
+.. autoclass:: QuadratureResidualJacobian
+.. autofunction:: quad_residual_and_jacobian
+.. autofunction:: quad_gauss_newton_increment
+"""
+
+from __future__ import annotations
+
+
+__copyright__ = """
+Copyright (C)  2024 University of Illinois Board of Trustees
+"""
+
+__license__ = """
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+"""
+import operator
+from collections.abc import Callable, Sequence
+from dataclasses import dataclass
+from functools import reduce
+from typing import TypeAlias
+from warnings import warn
+
+import numpy as np
+import numpy.linalg as la
+
+
+# FIXME: Better name?
+Integrand: TypeAlias = Callable[[np.ndarray], np.ndarray]
+
+
+@dataclass(frozen=True)
+class _ProductIntegrand:
+    functions: Sequence[Integrand]
+
+    def __call__(self, points: np.ndarray) -> np.ndarray:
+        return reduce(operator.mul, (f(points) for f in self.functions))
+
+
+@dataclass(frozen=True)
+class _ConjugateIntegrand:
+    function: Integrand
+
+    def __call__(self, points: np.ndarray) -> np.ndarray:
+        return self.function(points).conj()
+
+
+def _identity_integrand(points: np.ndarray) -> np.ndarray:
+    return points
+
+
+@dataclass(frozen=True)
+class LinearCombinationIntegrand:
+    """
+    .. note::
+
+        New linear combinations should be created by :func:`linearly_combine`,
+        which flattens nested combinations by taking advantage of associativity,
+        substantially decreasing evaluation cost.
+
+    .. autoattribute:: coefficients
+    .. autoattribute:: functions
+    .. automethod:: __call__
+    """
+    coefficients: np.ndarray
+    functions: Sequence[Integrand]
+
+    def __post_init__(self):
+        assert len(self.coefficients) == len(self.functions)
+
+    def __call__(self, points: np.ndarray) -> np.ndarray:
+        """Evaluate the linear combination of :attr:`functions` with
+        :attr:`coefficients` at *points*.
+        """
+        return sum(
+            (coeff * func(points)
+                for coeff, func in zip(self.coefficients, self.functions, strict=True)),
+            np.zeros(()))
+
+
+def linearly_combine(
+            coefficients: np.ndarray,
+            functions: Sequence[Integrand]
+        ) -> LinearCombinationIntegrand:
+    r"""
+    Returns a linear combination of *functions* with coefficients.
+    If *functions* are themselves :class:`LinearCombinationIntegrand`\ s,
+    and all *functions* use a consistent basis, then the resulting
+    :class:`LinearCombinationIntegrand` will be flattened via associativity,
+    leading to faster execution time.
+    """
+
+    lcfunctions = [
+        f for f in functions
+        if isinstance(f, LinearCombinationIntegrand)
+    ]
+    if len(lcfunctions) != len(functions):
+        if lcfunctions:
+            warn("Only some functions passed to linearly_combine were themselves "
+                 "linear combination. This case is unhandled."
+                 "Returning an un-flattened linear combination.",
+                 stacklevel=2)
+        return LinearCombinationIntegrand(coefficients, functions)
+
+    basis: list[Integrand] = []
+    n = len(lcfunctions)
+    matrix = np.zeros((n, n), dtype=np.complex128)
+    for i, f in enumerate(lcfunctions):
+        ncommon = min(len(basis), len(f.functions))
+        if basis[:ncommon]  != f.functions[:ncommon]:
+            warn("Not all functions passed to linearly_combine used the "
+                 "same basis. Returning an un-flattened linear combination instead.",
+                 stacklevel=2)
+
+            return LinearCombinationIntegrand(coefficients, functions)
+
+        basis.extend(f.functions[ncommon:])
+
+        ncoeff = len(f.coefficients)
+        matrix[i, :ncoeff] = f.coefficients
+
+    return LinearCombinationIntegrand(coefficients @ matrix, basis)
+
+
+def _mass_matrix(
+            integrate: Callable[[Integrand], np.inexact],
+            basis: Sequence[Integrand],
+        ) -> np.ndarray:
+    n = len(basis)
+    integrals = [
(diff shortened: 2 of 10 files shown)


inducer:
It made its first nodes:

grafik

:tada: 

cc @alexfikl

inducer:
One possible issue right now is the basis orthogonalization. V and R propose an SVD-based procedure that (IIUC?) seems to produce $\ell^2$ (rather than $L^2$)-orthogonal functions. I went with a Cholesky-based approach instead (somewhat analogous to CholeskyQR2, maybe), but it doesn't maintain orthogonality very well for larger (?) bases. For the order-7 example above, it only gets less than 12 digits.

alexfikl:
> I went with a Cholesky-based approach instead (somewhat analogous to CholeskyQR2, maybe), but it doesn't maintain orthogonality very well for larger (?) bases. For the order-7 example above, it only gets less than 12 digits.

Would a pivoted version work better? Although those aren't exactly available easily https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.lapack.dpstrf.html#scipy.linalg.lapack.dpstrf

inducer:
> Would a pivoted version work better?

TIL that's even a thing! In CS450, I say things like "Cholesky doesn't need pivoting". And it's [not just me](https://en.wikipedia.org/w/index.php?title=Cholesky_decomposition&oldid=1271259708):

> Now, suppose that the Cholesky decomposition is applicable. As mentioned above, the algorithm will be twice as fast. Furthermore, no [pivoting](https://en.wikipedia.org/wiki/Pivot_element) is necessary[...]

For now, I would bet that perhaps the bigger problem is that the implementation (just the test really) currently tries to go where it needs to by linearly combining monomials.

Another indication that it's perhaps not the linear algebra that is to blame here is that, at least by back-of-the-envelope analysis, the main "accuracy loser" in CholeskyQR2 is the formation of $A^TA$, which is not part of the method.

Yet another indication is that the coefficients reach a magnitude of about 3000, which might help explain the loss of about three digits.

inducer:
This is ready for a look. @shawnl00, while I can't formally request a review from you, I'd like you to review this code as well.

One thing that's missing is better support for (inefficiently) computing reference integrals of "obstinate" integrands, probably using adaptive quadrature. Some [useful pieces](https://documen.tician.de/modepy/shapes.html#modepy.submesh_for_shape) exist for that, but I figured I would still leave it for future work.

ShawnL00:
> This is ready for a look. @ShawnL00, while I can't formally request a review from you, I'd like you to review this code as well.
> 
> One thing that's missing is better support for (inefficiently) computing reference integrals of "obstinate" integrands, probably using adaptive quadrature. Some [useful pieces](https://documen.tician.de/modepy/shapes.html#modepy.submesh_for_shape) exist for that, but I figured I would still leave it for future work.

That is cool! will do!

modepy/quadrature/finding.py
```diff
@@ -0,0 +1,422 @@
...
+                for coeff, func in zip(self.coefficients, self.functions, strict=True)),
+            np.zeros(()))
+
+
+def linearly_combine(
+            coefficients: np.ndarray,
+            functions: Sequence[Integrand]
+        ) -> LinearCombinationIntegrand:
+    """
+    Takes advantage of associativity when linearly combining linear combinations.
```
alexfikl, line 111: This description should be a bit more thorough? From what I can tell, flattening works only if each linear combination has a subset of the same functions and if they're all in the same order?

modepy/quadrature/finding.py
```diff
@@ -0,0 +1,422 @@
...
+    Let :math:`\Omega\subset\mathbb C^n` be a domain. (Domains
+    over the reals are allowable as well.) Returns linear combinations
+    of functions in *basis* that is orthogonal under the (complex-valued)
+    :math:`L^2` inner product induced by *integrate*.
+
+    :arg integrate: Computes an integral of the passed integrand over
+        :math:`\Omega`. Must use integration nodes compatible with
+        *basis*.
+    :arg basis: A sequence of functions that accept an array of nodes
+        of shape either ``(ndim, nnodes)`` or ``(nnodes,)`` and return
```
alexfikl, line 170: Is `ndim` the n in $\mathbb{C}^n$?

modepy/quadrature/finding.py
```diff
@@ -0,0 +1,422 @@
...
+        an array of shape ``(nnodes,)`` with the value of the
+        basis function at the node.
+    """
+    n = len(basis)
+    mass_mat = _mass_matrix(integrate, basis)
+
+    l_factor = la.cholesky(mass_mat)
+
+    from scipy.linalg import solve_triangular
```
alexfikl, line 179: This should add some dependency on scipy in pyproject?

modepy/quadrature/finding.py
```diff
@@ -0,0 +1,422 @@
...
+        ) -> Sequence[Integrand]:
+    r"""Converts a list of :data:`Integrand`\ s taking nodes in real-valued
+    arrays of shape ``(2, nnodes)`` to ones accepting a single
+    complex-valued array of shape ``(nnodes,)``. See :func:`guess_nodes_vr`
+    for the main intended use case.
+    """
+    return [_ComplexToNDAdapter(f) for f in functions]
+
+
+def guess_nodes_vr(
```
alexfikl, line 210:
```suggestion
def vioreanu_rokhlin_nodes_guess(
```
?

modepy/quadrature/finding.py
```diff
@@ -0,0 +1,422 @@
...
+        a degree-1 monomial over :math:`\Omega`.
+    :arg onb: An orthonormal basis of functions. Each function takes
+        in an array of (complex) nodes and returns an array
+        of the same shape containing the function values.
+        Functions are expected to be orthogonal with respect to
+        the complex-valued inner product.
+
+        Integrands accepting real-valued node coordinates in two dimensions
+        (shaped ``(2, nnodes)`` may be converted to functions acceptable by
+        this function via :func:`adapt_2d_integrands_to_complex_arg`.)
```
alexfikl, line 232: Maybe worth adding a `is_real` flag or something to this to avoid exposing the complex adapter stuff?

modepy/quadrature/finding.py
```diff
@@ -0,0 +1,423 @@
...
+        of shape either ``(ndim, nnodes)`` or ``(nnodes,)`` and return
+        an array of shape ``(nnodes,)`` with the value of the
+        basis function at the node.
+    :arg nodes: shaped ``(ndim, nnodes)`` or ``(nnodes,)``, real-valued.
+        Must be compatible with *integrands*.
+    :arg reference_integrals: shaped ``(len(integrands),)``
+
+    .. note::
+
+        Tolerates overdetermined systems, will provide the least squares solution.
```
alexfikl, line 287:
```suggestion
        This function also supports overdetermined systems. In that case, it
        will provide the least squares solution.
```
? :grin:

modepy/quadrature/finding.py
```diff
@@ -0,0 +1,423 @@
...
+
+    if len(reference_integrals) != len(integrands):
+        raise ValueError(
+                "number of integrands must match number of reference integrals")
+    if len(reference_integrals) < len(nodes):
+        from warnings import warn
+        warn("Underdetermined quadrature system", stacklevel=2)
+
+    from modepy import vandermonde
+    return la.lstsq(vandermonde(integrands, nodes).T, reference_integrals)[0]
```
alexfikl, line 302: Does this fall back to a normal solve if the system is square?

alexfikl (approved):
Left a few nitpicks, but it all looks pretty straightforward. Very cool! :tada:

Some additional questions:
* How do the nodes from `guess_nodes_vr` compare to the stored VR nodes? Do they also miss some digits?
* Does `orthogonalize_basis` work to iteratively construct a basis too? I recall there was a section in Vioreanu2011 doing something like "start with f_i = 1, multiply it by x, orthogonalize, repeat".

modepy/quadrature/finding.py
inducer: I thought about this, but ultimately the inner product (i.e. the integration) fundamentally happens in the complex domain, and the integrands are allowed to evaluate to complex numbers (even if they don't for most use cases I have in mind). As a result, I figured that something is genuinely lost by plastering a pretend-real-valued interface over this. And, just as a matter of taste, I strongly prefer composition with an additional function over passing a vaguely opaque flag (easier to understand, for one, and not too much more typing).

I guess I could see a pretend-real *wrapper*? Let me try that out.

README.rst
```diff
@@ -4,9 +4,9 @@ modepy: Basis Functions, Node Sets, Quadratures
 .. image:: https://gitlab.tiker.net/inducer/modepy/badges/main/pipeline.svg
     :alt: Gitlab Build Status
     :target: https://gitlab.tiker.net/inducer/modepy/commits/main
-.. image:: https://github.com/inducer/modepy/workflows/CI/badge.svg?branch=main&event=push
+.. image:: https://github.com/inducer/modepy/actions/workflows/ci.yml/badge.svg
```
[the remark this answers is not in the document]
inducer: Yep.

modepy/quadrature/finding.py
inducer: Grr, you're right. And: no, I'm not doing that. I'd rather fall back to `la.solve`, even if it's asymptotically wrong.

modepy/quadrature/finding.py
inducer: Numerically, it's equivalent, in the square, full-rank case. Cost-wise, you're still paying for the SVD.

inducer: (I don't care much about the square use case here: If someone is using this, they're likely aiming higher than Newton-Cotes.)

doc/quad_finding.rst
```diff
@@ -0,0 +1,4 @@
+Support for development of new quadrature rules
+===============================================
+
+.. automodule:: modepy.quadrature.finding
```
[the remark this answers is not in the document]
inducer: I like "construction". Let's go with that.

modepy/quadrature/finding.py
inducer: Fixed.

modepy/quadrature/finding.py
inducer: Done.

inducer:
> * How do the nodes from `guess_nodes_vr` compare to the stored VR nodes? Do they also miss some digits?

Not a meaningful comparison IMO. `guess_nodes_vr` gives interpolatory nodes. Off the bat, the nodes out of `guess_nodes_vr` only integrate $P^N$, nothing more, at least in the test code. The stored VR nodes are the result of additional "tightening" via Gauss-Newton. And our Gauss-Newton widget isn't yet full-featured enough to compare. For instance, if I saw this correctly, the nodes lose symmetry pretty much right away. It does manage to turn nodes for $P^7$ (36 quadrature nodes) into ones that integrate $P^{11}$ (78 functions $\approx 36\cdot2.166$. By way of conjecture, 108=36*3 integrands should be attainable.).

> * Does `orthogonalize_basis` work to iteratively construct a basis too? I recall there was a section in Vioreanu2011 doing something like "start with f_i = 1, multiply it by x, orthogonalize, repeat".

TBH, I did not fully understand that bit of the paper. As it stands, it seems that this process would provide functions that are orthogonal in terms of the $\ell^2$ norm applied to the nodal values, not the $L^2$ sense that seems mathematically needed. It also seems require representing the integrands as a bunch of nodal values, which doesn't leave an obvious way to find basis derivatives that are needed for the Gauss-Newton process. So, in all likelihood, I'm still misunderstanding something.

> * How slow is this to find higher order rules?

The tests run within a few seconds, but they only exercise up to $P^{11}$ on the triangle. So it could probably be sped up, but I'll wait with that until it's actually too slow in a specific use case. I haven't profiled, but I suspect the evaluation of the PKDOs is the bottleneck.

modepy/quadrature/finding.py
inducer: I've ended up with `guess_nodes_vioreanu_rokhlin`. I find that "lead with a verb" is a good guideline for function names.

modepy/quadrature/finding.py
inducer: Tried the wrapper. Not worth it IMO. To be useful, it would have to deal with `(nnodes,)` and `(2, nnodes)`, and this caused enough grief to make me want to run away. So nah, not happening, sorry!

inducer:
Pushed the changes addressing the concerns.

modepy/quadrature/finding.py
alexfikl: Thanks! Looks good!

modepy/quadrature/finding.py
alexfikl: I was thinking of guess as a noun :grin: Looks good!

modepy/quadrature/construction.py
```diff
@@ -0,0 +1,451 @@
...
+    :func:`guess_nodes_vioreanu_rokhlin`,
+    and an initial guess for the weights (given nodes) may be found
+    via :func:`find_weights_undetermined_coefficients`.
+
+    .. note::
+
+        Unlike :func:`guess_nodes_vioreanu_rokhlin`, domains of any dimensionality
+        are allowed.
+
+    :arg nodes: shaped ``(ndim, nnodes)`` or ``(nnodes,)``, real-valued.
```
alexfikl, line 387:
```suggestion
    :arg nodes: An array with shape ``(ndim, nnodes)`` or ``(nnodes,)``, real-valued.
```

modepy/quadrature/construction.py
```diff
@@ -0,0 +1,451 @@
...
+        Must be compatible with *integrands*.
+    :arg weights: shaped ``(nnodes,)``, real-valued
```
alexfikl, line 389:
```suggestion
    :arg weights: An array with shape ``(nnodes,)``, real-valued
```

alexfikl (approved):
Left a few nitpicks in the docs to make them sound more sentence-y, but otherwise this looks good to me! :heart:

inducer:
Forgot the most important part: Advertise the existence of this in the README! :grin: 3dbe7d3be357ac04b23996e99baebbb52122f948

PR #109, opened 2025-02-08
Add script to extract Womersley spherical quadrature rules

alexfikl:
This adds a little script to extract the quadrature rules from Womersley (2018) (https://doi.org/10.1007/978-3-319-72456-0_57). They can be found at https://web.maths.unsw.edu.au/~rsw/Sphere/EffSphDes/sf.html and https://web.maths.unsw.edu.au/~rsw/Sphere/EffSphDes/ss.html.

Unfortunately, the rules are quite huge (up to 52978 points), so we can't really include them in the code. I tried making a little file like the existing `xg_quad_data.py` and friends and it was about 400MB uncompressed.. The current script saves the extracted rules in an `npz` file that's about 90MB.

Not sure if this is useful, but I'll leave it here :grin:

### DIFF
--- a/contrib/weights-from-womersley.py
+++ b/contrib/weights-from-womersley.py
@@ -1,0 +1,161 @@
+"""
+The associated paper can be found at
+
+    R. S. Womersley,
+    *Efficient Spherical Designs With Good Geometric Properties*,
+    Springer International Publishing, pp. 1243--1285, 2018,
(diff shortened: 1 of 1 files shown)


inducer:
Thanks! IMO we could merge this, but I won't as long as you have it as draft.

alexfikl:
> Thanks! IMO we could merge this, but I won't as long as you have it as draft.

I'm definitely fine with merging it too. :grin:

PR #123, opened 2025-07-13
Improve space typing and __new__ overloads

alexfikl:
Follow-up to #122 for `FunctionSpace.__new__`. This should fix the typing issues in #121 :grin:


inducer:
Thx!

PR #81, opened 2024-04-29
Fix support for zero-dimensional tensor product discretizations

inducer:
- [x] Add tests

PR #127, opened 2025-07-18
Port plot-basis to modern modepy

alexfikl:
The `plot-basis.py` example was broken when `simplex_onb` was removed in #81. This updates the example to use the new API.


inducer:
Thx!

PR #128, opened 2025-07-24
Clean up quad example and add more nodes

alexfikl:
It now looks a bit like below: made it blue, added vertices, removed the axis. Mostly modified it to make some prettier pictures for the paper.

quadrature_rule_gm_order_11_2d


inducer:
Thx!

PR #129, opened 2025-07-25
Deprecate automatic flattening for tensor products

### LINKED ISSUE #126: Stop auto-flattening and simplifying TensorProducts
This happens for `TensorProductShape` and `TensorProductSpace`. Mainly, we should remove
* The simplification in `__new__` for "tensor products" with a single object.
* The flattening in `__init__`.

Motivation: just tried using the code below and it got flattened to 6 PN spaces, which doesn't seem desired.
```python
space = mp.TensorProductSpace((mp.QN(3, 5), mp.QN(1, 5), mp.QN(2, 5)))
assert len(space.bases) == 6
```

Thoughts?

alexfikl:
This adds a `.flatten()` method to tensor products to recursively flatten them into simplices (mostly). The automatic flattening in the constructor is marked as deprecated.

Fixes #126.


inducer:
I'm supportive of this change. How come it's marked as draft?

PR #130, opened 2025-07-29
Add sphinx helper to reroute references

alexfikl:
This adds a little helper to the sphinx config to re-route missing references and remove the need to add dummy entries to the actual docs.

- [x] Needs https://github.com/inducer/sphinxconfig/pull/1. 
- [x] Switch link to `sphinxconfig.py` after that is in.


doc/conf.py
```diff
@@ -24,9 +24,58 @@
...
+    # modepy typing
+    "ArrayF": (None, "modepy.typing", "obj"),
+    "BasisFunction": (None, "modepy.typing", "obj"),
+    "BasisGradient": (None, "modepy.typing", "obj"),
+    "NodalFunction": (None, "modepy.typing", "obj"),
+    "RealValueT": (None, "modepy.typing", "obj"),
 }
+
+
+def process_autodoc_missing_reference(app, env, node, contnode):
```
inducer, line 45: Annotate?

doc/conf.py
```diff
@@ -24,9 +24,58 @@
...
 autodoc_member_order = "bysource"
 
-# FIXME: This would be nice to have, but I was not able to resolve the
-# resulting mess of Sphinx errors.
-#
-autodoc_type_aliases = {
-    # "ArrayF": "modepy.typing.ArrayF",
+
+# NOTE: values are `(inventory, module, reftype)`
+custom_type_links = {
```
inducer, line 29:
```suggestion
custom_type_links: dict[str, tuple[str | None, str, str]] = {
```

inducer (commented):
Dang, this is awesome! :grin: Thanks for working on this. Having spent lots of unnecessary time adding this nonsense, this feels like a godsent, so much so that I feel that it should live in https://github.com/inducer/sphinxconfig/blob/main/sphinxconfig.py. Move there?

PR #132, opened 2025-08-03
Add type annotations to modepy.tools

alexfikl:
Mostly wanted `equilateral_to_unit` for some stuff in `meshmode`, but added more types to the whole module. As usual, the main leftovers are due to `ndarray.reshape` and friends.

### DIFF
--- a/modepy/tools.py
+++ b/modepy/tools.py
@@ -57,13 +57,12 @@ THE SOFTWARE.
 
 import math
 from typing import TYPE_CHECKING, Protocol, TypeVar, runtime_checkable
-from warnings import warn
 
 import numpy as np
 import numpy.linalg as la
 from typing_extensions import override
 
-from pytools import MovedFunctionDeprecationWrapper, memoize_method
+from pytools import memoize_method
 
 import modepy.shapes as shp
 from modepy.spaces import FunctionSpace, TensorProductSpace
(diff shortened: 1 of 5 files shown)


modepy/tools.py
```diff
@@ -73,70 +70,82 @@
...
+
+    from numpy.typing import ArrayLike
+
     from modepy.shapes import Shape
     from modepy.typing import ArrayF
 
 
+# NOTE: some places imported gamma from here, but this is deprecated and does
+# not seem to be used anywhere
+gamma = math.gamma
```
alexfikl, line 83: Can this be removed? I couldn't find any place that imported `gamma` from here and don't recall when that was even the case :\

modepy/tools.py
```diff
@@ -181,78 +193,81 @@ def inverse(self):
...
 
-def unit_vertices(dim):
-    warn("unit_vertices is deprecated. "
-            "Use modepy.unit_vertices_for_shape instead. "
-            "unit_vertices will go away in 2022.",
-            DeprecationWarning, stacklevel=2)
+def unit_vertices(dim: int) -> ArrayF:
+    warn("'unit_vertices' is deprecated. Use 'mp.unit_vertices_for_shape' instead. "
+         "'unit_vertices' will go away in 2022.",
+         DeprecationWarning, stacklevel=2)
```
alexfikl, line 214: There's this and the `submesh` functions that have been deprecated for a long time. Remove?

inducer: Yeah, kill them.

Another good (new-ish) option is to slap a `@typing_extensions.deprecated` decorator on there. That will start failing downstream bpr CIs (without impacting any code functionality).

inducer (commented):
Thx!

PR #134, opened 2025-09-02
JOSS paper

alexfikl:
This adds the JOSS paper to the `doc` folder and includes a CI action to build it.

**NOTE**: This PR will not be merged! It is a place to discuss the JOSS paper and make modifications during review.

### DIFF
--- a/doc/paper/paper.bib
+++ b/doc/paper/paper.bib
@@ -250,6 +250,7 @@
   year          = 2017,
   journal       = {PeerJ Computer Science},
   publisher     = {PeerJ Inc.},
+  doi           = {10.7717/peerj-cs.103},
   volume        = 3,
   pages         = {e103},
 }
--- a/doc/paper/paper.md
+++ b/doc/paper/paper.md
@@ -125,10 +125,10 @@ space of functions $\phi_i: D \to \mathbb{R}$, where $D$ is the reference elemen
 domain, and no specific choice of basis. Predefined choices include the `PN` space,
 containing polynomials of total degree at most $N$, and the `QN` space, containing
 polynomials of maximum degree at most $N$. As with shapes, these spaces can be
-combined using `TensorProductSpace`. A `Basis` objects is available separately,
+combined using `TensorProductSpace`. A `Basis` object is available separately,
 giving access to basis functions and their derivatives, for, e.g., the monomials,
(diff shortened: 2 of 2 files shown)


inducer:
Thanks for getting this finished up! That said, I'm not sure the modepy repo is the right spot for it. I don't intend to "maintain" the paper in the same way I intend to maintain the docs or the code. Once it's finished, it's finished in my opinion, and the archival copy will live at JOSS. We can link to it, but IMO it's not great to include it here.

alexfikl:
> Thanks for getting this finished up! That said, I'm not sure the modepy repo is the right spot for it. I don't intend to "maintain" the paper in the same way I intend to maintain the docs or the code. Once it's finished, it's finished in my opinion, and the archival copy will live at JOSS. We can link to it, but IMO it's not great to include it here.

My understanding of their submission documentation (from https://joss.readthedocs.io/en/latest/submitting.html#submission-process) is that the paper should be in the same repository somewhere. The submission form also has a `Name of git branch containing the paper` field, so it seems like they expect it to be somewhere in the repo.

We could make this a "branch" with no other history and just leave it there. Would that be ok? I mostly just made this PR to make sure that the build works, but I'm fine with not actually merging it into `main`.

inducer:
Reading their submission guidelines, I don't see anywhere that explicitly says that the paper has to be in the same repo as the code. And I'm not a huge fan of branch-is-separate-thing, it feels like that's holding git wrong.

My read of the guidelines is that they want *a* repo (which IMO makes sense for change tracking) but aren't particular about which. As such, I think I'd be happiest if we could just move https://gitlab.tiker.net/papers/2025-modepy-joss someplace public.

What do you think?

inducer:
A quick random sample of other JOSS submissions seems to indicate that that's the common approach.

inducer:
cc @a-alveyblanc @thomasgibson @xywei for visibility.

alexfikl:
> Huh, a closer look at the submission form doesn't seem to allow for the two-repo solution. So, fair, let's let this branch be the authoritative source for the paper, but let's not plan on merging it. Converting to draft to that end.

Yeah, the submission guidelines just sort of implicitly assume that it's the same repo (by mostly not talking about two repos in any way), but the submission form seems pretty clear cut that it wants a branch in the code's repo.

Marking this a draft and not merging it sounds good to me!

PR #136, opened 2025-09-04
Remove non-standard text from LICENSE

alexfikl:
Recommended in https://github.com/openjournals/joss-reviews/issues/8926#issuecomment-3250926190.


inducer:
Thx!

PR #144, opened 2025-10-19
Small type updates

alexfikl:
https://github.com/inducer/modepy/actions/runs/18624857390/job/53101501661

Most of those failures seem to be related to numpy 2.3.4 (?), so they just went into the baseline.


inducer:
Ty!

PR #164, opened 2026-03-16
Add type annotations to tests

alexfikl:
This mostly adds more type annotations to the tests.

The only interesting change is the update to `Reshapeable`. I changed it to at least accept `ArrayF` now. From a quick grep, this seems to only be used in `meshmode` (and this change makes the cast unnecessary):

https://github.com/inducer/meshmode/blob/6bbb682185aadd4f299f851cdfe5805fdb4c8f1a/meshmode/discretization/poly_element.py#L567-L569

### DIFF
--- a/.basedpyright/baseline.json
+++ b/.basedpyright/baseline.json
@@ -1650,62 +1650,6 @@
                 }
             },
             {
-                "code": "reportUnknownParameterType",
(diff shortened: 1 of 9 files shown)


inducer:
Thx!

PR #165, opened 2026-03-19
Order vector tp

inducer:
- [x] Still waiting on a test, currently being made by Claude (maybe?)

cc @shawnl00

This makes https://github.com/inducer/meshmode/pull/494 way neater.

### DIFF
--- a/modepy/modal_decay.py
+++ b/modepy/modal_decay.py
@@ -23,14 +23,18 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
 """
 
-
+from functools import singledispatch
 from typing import TYPE_CHECKING
 
 import numpy as np
 import numpy.linalg as la
+from typing_extensions import deprecated
+
+from modepy.modes import Basis, SimplexBasis, TensorProductBasis
 
 
 if TYPE_CHECKING:
+    import optype.numpy as onp
     from numpy.typing import NDArray
 
     from modepy.typing import ArrayF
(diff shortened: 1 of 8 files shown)


modepy/modal_decay.py
```diff
@@ -49,25 +52,82 @@
...
+def degree_vector_tp(basis: TensorProductBasis) -> ArrayF:
+    degree_vectors = [degree_vector(b) for b in basis.bases]
+    return np.array([
+                        max(
+                            ov[mode_id]
+                            for mode_id, ov
+                            in zip(mid_tuple, degree_vectors, strict=True)
+                        )
+                        for mid_tuple in basis._mode_index_tuples  # pyright: ignore[reportPrivateUsage]
+                    ])
```
alexfikl, line 92: I think I would prefer to have an `if isinstance(...): ...` thing inside the function instead? I don't imagine these will be used anywhere else.

Alternatively, maybe it's better to add them to the `Basis` class next to `mode_ids`?

alexfikl (approved):
Left a little nitpick, otherwise looks good to me!

modepy/modal_decay.py
inducer: I don't like adding method clutter, this doesn't seem like it's part of the minimal set of things that the class needs to offer.

And I don't view `isinstance` as preferable; I don't see any advantages. Could you explain?

alexfikl: > I don't like adding method clutter, this doesn't seem like it's part of the minimal set of things that the class needs to offer.

Yeah, agreed.

> And I don't view isinstance as preferable; I don't see any advantages. Could you explain?

This seems like it's just `isinstance` with extra steps, since it's not an interface that will likely be extended or used in other places. 

Either way, I don't feel strongly about it, since it's not a hot path or anything. It just seemed like a lot more machinery than I expected.

inducer: As a counterpoint, I think it's as likely to be extended as other places in modepy. :slightly_smiling_face: It's reversible, too, without breaking compat, so it probably doesn't matter too much.

It's not that I don't appreciate your opinion! It's just that I disagree today :grin:.

alexfikl: @inducer Slightly related, does this work on like a prism, where the tensor product is between a `QN` and a `PN` basis?

alexfikl: Ah, just saw that it does take the degrees from the component bases. Nevermind!

inducer: Yep! That was my whole motivation behind pushing this into modepy, since it actually makes use of its structure. :grin:

PR #166, opened 2026-03-21
Fix up docs for transplanted quadrature

alexfikl:
This mostly deletes the text from `quadrature.rst` because it just repeats what's in the function docs, e.g.
https://documen.tician.de/modepy/quadrature.html#strip-conformal-map


inducer:
Thx! I should've caught that. Oh well. Thanks!

